PR #13, opened 2019-09-04
65C02 Opcode upgrade

paulscottrobson:
65C02 fixes:

- split fake6502.c into subcomponents
- autogenerates tables, disassembly (hence disasm.c changed)
- seperate file with 65C02 new stuff
- small changes to 6502 for 65C02 compatibility (see README in cpu)


sebastianvog:
chargen.bin and rom.bin should not be committed, or ? Perhaps those two files should get added to .gitignore so it is not accidently part of a commit.

mist64:
Also, cpu/tables.h.org doesn't seem it should be part of this.

mist64:
I fixed the issues and merged it as ac46b2a.

mist64:
P.S.: Thanks Paul! :)

PR #15, opened 2019-09-05
ASCII->PRG converter in Python

paulscottrobson:
One Python script, no support files needed, but has to be updated manually. Apologies for the re-pull. Documentation on fork syncing is .... variable shall we say :)


mist64:
Instead of having an external tool do the work, which will have to be updated with changes in BASIC – and BASIC will change a lot! – I prefer reusing the existing 6502 code. I have added logic to allow pasting ASCII into the emulator, and to echo screen output to the host's stdout.

What do you think?

paulscottrobson:
Have both :) 

Or I can change it so it auto-rips from the ROM image. There's advantages to both. This way you need one file.

paulscottrobson:
First working debugger version. Seems to work perfectly fine.  The only 'problem' is the keys. You can't single step through key code, because it's designed on a minimum-changes methodology and I don't want to tinker with stuff that works. So when it's debugging (not when you are running to a breakpoint, or just running, or stepping over a subroutine) it grabs the SDL Event queue for itself. Press F5/Run and the emulator has it as normal.

Despite the 13 files, it actually is two new files, debugger.c debugger.h . Code is injected at two points, one at the top of the main execution loop in main.c, and the other just before rendering in video.c. "debugger.h" needs to be included in both. Still battling with multi-user github :)

Full details are in debugger.c. I would suggest having a -debug option that starts it in debug mode (as per the notes in the file), which will mean it is effectively transparent without it.

Keys are based on Microsoft's standard ones (F10-Step Over, F11-Run etc.) and easily changeable anyway.

mist64:
This looks awesome! Can you create a new (clean) pull request? :)

PR #19, opened 2019-09-05
Clean Debugger

paulscottrobson:
Clean version of the debugger. This is one that autostarts it, which would be annoying, so I suggest adding a -debug option as per the code in debugger.c. If you never call DEBUGGetCurrentStatus() it should have no effect whatsoever.


mist64:
Merged. Thanks a lot! Can you add some documentation to README.txt?

PR #20, opened 2019-09-06
README.md / Debugger

paulscottrobson:
Have added a short paragraph on the debugger. Ignore the other 2 files, still battling with git :(
You may want to look at the changes in video.c which is using scancodes not SDL key symbols. I understand SDL key symbols tinker about in an attempt to be helpful, whereas I believe we want the 'raw' PS/2 Scancodes as close as possible.


mist64:
git is tricky. I think you need to create a new branch from my current master and cherry-pick your change on top of it, then all I get is the single change that I can apply with a button.

mist64:
cde23c9

PR #22, opened 2019-09-06
Debugger tweaks

paulscottrobson:
First : Two changes, neither huge.

1) Instead of starting in STOP mode, the debugger starts in RUN mode. It's unlikely most developers will be wanting to single step from Boot.
2) The TAB key now hides the debug window while it is pressed, so it is possible to see the whole display.

Second : Not huge but slightly more complex

Added pseudo operation '$FF' "dbg" which when executed breaks into the debugger. Allows insertion of automatic debuggers in code.


mist64:
Great!

In future patches, can you make sure the coding style matches the rest of the code better? E.g.

* Always { } after `if`.
* Tabs only for code indentation; spaces only for aligning comments

paulscottrobson:
Sorry, force of habit.  Not sure I can get SublimeText to do tabs for indent and spaces for code.

PR #26, opened 2019-09-07
Allow emulation to switch emulator features on or off.

rsbohn:
This adds a new device at $9FB0 to $9FB4 that is wired
to some of the emulator features. Allows you to turn features
on and off from within the emulation.
$9FB0	Debugger
$9FB1	Video access logging
$9FB2	Keyboard logging
$9FB3	Echo mode
$9FB4	Save on exit:
	Should the emulator save a memory image
	when $FFFF is called?

Usage:
POKE $9FB3,1:PRINT"ECHO MODE IS ON":POKE $9FB3,0

POKE $9FB4,0:SYS $FFFF:REM Exit without saving a memory image.

### DIFF
--- a/glue.h
+++ b/glue.h
@@ -22,6 +22,8 @@ extern uint8_t ROM[];
 extern bool debuger_enabled;
 extern bool log_video;
 extern bool log_keyboard;
+extern bool echo_mode;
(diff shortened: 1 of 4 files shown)


mist64:
Great!

Issue #33, opened 2019-09-08
Load kernel call doesn't load into the currently set RAM bank


mist64:
[load-page-issue-emu-29.zip](https://github.com/commanderx16/x16-emulator/files/3587970/load-page-issue-emu-29.zip)

Program to demonstrate loading issue into page $A000. It appears as though the load
kernel call doesn't load into the currently set ram bank. 

This program loads demo data into page 0 and 1, prints it out, then loads a file
(while bank $1 is active), after the load the bank data is printed out. 

1) Copy loadpage.prg and page1.dat to the emulator directry
2) Cold boot emulator 29
3) Load the example program
   LOAD "loadpage.prg",1,1
4) Run the program
   SYS 4096
*) Loadpage produces the following output

PAGE 0 BEFORE LOAD: PAGE0
PAGE 1 BEFORE LOAD: PAGE1
-------------------------------------
LOADING PAGE1.DAT
CURRENT BANK IS: 1
-------------------------------------
PAGE 0 AFTER LOAD : DATA FILE DATA FILE DATA FILE
PAGE 1 AFTER LOAD : PAGE 1

*) The string "DATA FILE", etal, should be in bank 1 and page 0 should display "PAGE 0"

Bonus behavior
5) Run test program again
   SYS 4096

*) The first line now displays
   PAGE 0 AFTER LOAD : DATA FILE DATA FILE DATA FILE
   (it should have displayed : PAGE0)

   It almost looks like writes to bank 0 were ignored.

SDCARD test
6) Modify device number in loadpage.a to 8 (in SETLFS)
7) Re-assemble loadpage.prg
8) Copy loadpage.prg and page1.dat to sd card
9) Cold boot emulator
10) Load program
    LOAD "LOADPAGE.PRG",8,1
11) Run
    SYS 4096
*) Same output behavior as previously noted on device 1 and Bonus output.

indigodarkwolf:
Thanks for accepting my fix!

mist64:
Thanks for fixing! :D

PR #34, opened 2019-09-09
Fix to issue #33: LOAD into RAM bank

indigodarkwolf:
Adding logic to LOAD() to place the file into the proper RAM bank if loaded into a RAM bank address. I'm not sure if final hardware/kernel will intelligently cap read length to the banked memory limit, similar to how the emulator enforces the fixed memory limit, or if large files would potentially overflow (failing to write to ROM, of course) and/or wrap-around back to RAM address 0x0000 and write forwards from there.

I've assumed an intelligent cap will be enforced like in the fixed RAM case. If that's not the case, I'm happy to instead implement whatever the behavior should be.

### DIFF
--- a/loadsave.c
+++ b/loadsave.c
@@ -9,6 +9,7 @@
 #include <dirent.h>
 #include <unistd.h>
 #include "glue.h"
+#include "memory.h"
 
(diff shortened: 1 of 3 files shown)


mist64:
Great!

PR #36, opened 2019-09-10
Basic examples

daitangio:
A set of basic examples found on the Facebook page. Added also a classic "maze" example, taking advantage of the 80 columns display


indigodarkwolf:
Should these examples go to https://github.com/commanderx16/x16-demo, instead?

mobluse:
I think there should be a separate repository for example X16 BASIC programs in text form, i.e. programs that can be loaded with ./x16emu -bas path/to/file.bas

indigodarkwolf:
Or maybe move the assembly examples into a subdirectory of the x16-demo repo, and the BASIC programs live in a different subdirectory?

daitangio:
Ok, I got it :)
I will make a pull request soon, okey?

Please excuse typs as Sent from my iPhone

indigodarkwolf:
Looks like these got pulled into https://github.com/commanderx16/x16-demo, can we close this?

PR #60, opened 2019-09-15
Gif recording capability

neilogd:
Make recording to a GIF trivial by commandlike "-gif test.gif"


neilogd:
Might do a little more refinement of this, does work, but the animation delay seems to be a bit much.

mist64:
Awesome. Can you try to make the coding style (of the changes; I'm not talking about gif.h) a little more consistent with the existing code though, please? It's basically BSD style.

* asterisk with the variable, not the type.
* open brace on the same line
* space after keywords like `if`
* always braces after `if`

Thanks!

neilogd:
Should be sorted now. Frame rate of the recorded gif still doesn't seem right. In the past when embedding this in projects I've added a frame skip function to avoid exporting as many frames, and achieve better playback speed. Some image viewers seem to respect the 10ms frame delay, where as browsers seem to not, frustratingly.

neilogd:
Aaand there we are, a nice to look at frame rate: https://twitter.com/neilogd/status/1173186178062749696

Kroc:
Suggestion: APNG is an easy format to write and would be smaller in practice than GIF.

mist64:
Great work, great feature. Thanks.

PR #61, opened 2019-09-15
Adding support for HTML5/WebAssembly Build Target

sebastianvog:
This Pull Request adds support for HTML5/WebAssembly compilation using the Emscripten compiler.
I tried to make the changes to existing code minimal/isolated. A demo from the output is here:
https://sebastianvog.github.io/x16-emulator/x16emu.html (Chrome only)

From a structural change perspective, I moved the emulator into its own emulator_loop. The way the code is executed won't change for a normal make/build target. It should work as before.

When compiling with emmake,  the main emulator loop will run in its own thread, and we are using the emscripten browser call back to update the screen at 60fps. This keeps the browser responsive.

From the WebAssembly.md:
## WebAssembly/HTML5 Support
The Commander X16 emulator supports a HTML/WebAssembly build target using the [Emscripten compiler](https://emscripten.org/).

## Demo
[HTML X16 Emulator Demo](https://sebastianvog.github.io/x16-emulator/x16emu.html)
## Installation
Follow installation steps from [https://emscripten.org/docs/getting_started/downloads.html#platform-notes-installation-instructions-sdk](https://emscripten.org/docs/getting_started/downloads.html#platform-notes-installation-instructions-sdk)

## Build

    emmake make

This outputs the following artifacts in the build directory, which can be uploaded to any web server.

	ex16mu.data x16emu.html x16emu.js x16mu.wasm x16emuworker.js

### Testing
To run a test webserver:

	python -m SimpleHTTPServer 8080

And start [http://localhost:8080/x16emu.html](http://localhost:8080/x16emu.html)

### Supported Browsers
Currently supported Browser is Google Chrome only.

FireFox will work if SharedMemory is enabled from preferences (but comes with Security implications).

Safari is not supported.


kktos:
Absolutly brilliant ! Will give it a try. Cheers to you.

mobluse:
What is the purpose of the box below the emulator? I tried to paste into it, but nothing happened in the emulator. Maybe it's for echo. 

I SAVEd a program and did new and LOADed it again, but where are the programs stored?

sebastianvog:
> What is the purpose of the box below the emulator? I tried to paste into it, but nothing happened in the emulator. Maybe it's for echo.
> 
> I SAVEd a program and did new and LOADed it again, but where are the programs stored?

The bottom box is for standard output from the emulator (like printf statements in the code), eventually this pane probably can be hidden/toggled. Nothing persists other than in the x16 emulator during the run. If you reload the webpage everything will be reset. So I am a bit surprised that you say you saved a program, did a new and it loaded again. Could explain a bit more on this?

There is no way for the web version to load any programs in yet or pass any parameters in. I hope to work on this soon, so one could load code or prg s  into the emulator.

mobluse:
Maybe you could support Ctrl+V for pasting in programs (this works in the desktop version). One solution might be to have a box you paste into and then it gets pasted into the emulator. It would be interesting to see how fast it is. Ctrl+S works for saving and that also prints in the box. Ctrl+R for reset also reloads the web page. Ctrl+Return doesn't work, but Ctrl+F does (after you close the find box).

sebastianvog:
Yes good idea. It would be very useful if Ctrl-V would work.  I noticed it didn't, but I have not had the time to look into it. it looks like sdl/emscripten doesn't support reading out the clipboard. I should be able to callback to javascript and read the clipboard out ( or a text box) and send back the results to the web assembly code.

richardwooding:
This is awesome, hopefully your PR will be approved.

antmurph:
For testing you don't necessarily need to use Python, you can use 'emrun x16emu.html' same idea

sebastianvog:
I added very basic support for getting a clipboard from the HTML side into the emulator. Not very elegant, but better than nothing I suppose.

I updated https://sebastianvog.github.io/x16-emulator/x16emu.html

mist64:
Sorry for the delay, I'll have a look soon!

sebastianvog:
I had some time for general cleanup, updated clipboard support via HTML, added reset button, and also got the latest from master to be up-todate.
Updated https://sebastianvog.github.io/x16-emulator/x16emu.html

tysepa:
yes added reset button

mist64:
I fixed the conflicts, squashed the commits, fixed a trivial compile issue on Linux and merged it to master: d96785c

mist64:
P.S.: GREAT WORK! :D

mist64:
P.P.S.: What's the deal with Shared Memory? Why does it not work in Safari? Any architectural changes to the emulator that would make it more compatible with emscripten's model?

sebastianvog:
> P.P.S.: What's the deal with Shared Memory? Why does it not work in Safari? Any architectural changes to the emulator that would make it more compatible with emscripten's model?

I have been using pthreads for the main emulator loop to run the emulator code as-is but also keep the browser responsive. pthreads in emscripten rely on SharedMemoryBuffer, which was removed in all browsers 2018 due to the Spectre vulnerability. 
https://www.sitepen.com/blog/the-return-of-sharedarraybuffers-and-atomics/ 
Chrome was able to bring them back, on FireFox you can configure a setting to allow them.  I think SharedArrayBuffer is the only issue in Safari, although I am not 100%. 
 An option would be to remove the thread and do the emulation_loop iteratively and yield back execution to the browser.I am hoping to explore this a bit more. It would be nice if Safari and FireFox would work right out of the box. Also performance certainly lags behind the native compiled version, so perhaps not using threads would make things faster. I will give it a try

PR #80, opened 2019-09-18
first implementation for YM2151 emulation

Frank-Buss:
Might be interesting for the emulator, if the YM2151 is going to be part of the Commander X16 hardware. Only problem is that the YM2151 implementation is from this project:
https://github.com/lantus/Cannonball-C/tree/master/src/main/hwaudio
This project doesn't have any license information, but looks like the source code is from the MAME project, which I think is GPL licensed. This would be in conflict with the BSD license, as discussed on Facebook.

But maybe it is not a problem. The Commander X16 emulator would become GPL licensed, but it might be possible not to get infected by GPL: I think you would need only to remove the YM2151 files, and then it would be BSD licensed again. Or to say it differently: The Commander X16 emulator without the YM2151 files stays BSD licensed, but only for each release it would be combined with the YM2151 drivers, which would generate a GPL version. Could be technically encoded in the Makefile as well. This would allow to use it in other commercially projects, if the YM2151 driver is not used. Should be clarified in a license description file. It is not uncommon for project to have multiple licenses.

But I'm not a lawyer, and honestly I don't care much, but might be important for David that the whole emulator project doesn't get "GPL infected" because of this driver. The audio part itself which I implemented in the main.c file is of course BSD licensed, and can be later enhanced by the DAC chip, if needed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,8 +24,8 @@ ifeq ($(CROSS_COMPILE_WINDOWS),1)
 	CC=i686-w64-mingw32-gcc
 endif
 
-OBJS = cpu/fake6502.o memory.o disasm.o video.o ps2.o via.o loadsave.o spi.o vera_spi.o sdcard.o main.o debugger.o
-HEADERS = disasm.h cpu/fake6502.h glue.h memory.h video.h ps2.h via.h loadsave.h
+OBJS = cpu/fake6502.o memory.o disasm.o video.o ps2.o via.o loadsave.o spi.o vera_spi.o sdcard.o main.o debugger.o ym2151.o
+HEADERS = disasm.h cpu/fake6502.h glue.h memory.h video.h ps2.h via.h loadsave.h ym2151.h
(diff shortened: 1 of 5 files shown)


mist64:
I don't want to switch to the GPL, but I am not opposed to merging this either. As long as the GPL part is a compile time option, I'm all for it. Can you make it an option?

Frank-Buss:
Good idea. I modified it. If you set a variable like this:
```
export WITH_YM2151=1
```
then "make" will compile it with the emulation. Test program:
http://www.frank-buss.de/x16/audio.prg

mist64:
Thanks. I do not intend to enable YM2151 for release binaries though: As far as I understand, no final decision has been made on the question of the sound chip, so adding one to the emulator binaries is skewing the discussion. Having to compile the emulator yourself (or finding an alternate source for a binary) is enough of an obstacle though, in my opinion.

PR #83, opened 2019-09-19
Compile for Linux and macOS on Travis CI

tobier:
This patch introduces continuous integration for Linux and macOS using Travis CI. It should help verifying that pull requests actually compile and for multiple platforms. Even better, if test cases are added then they can be run automatically to verify nothing breaks.

A future improvement to this would be to add AppVeyor that supports Windows builds.

**Note** that the owner of the repository will have to go to Travis-CI.org and enable CI for this repository. I've done it for my fork and both Linux and macOS builds pass.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,20 @@
+dist: bionic
+sudo: required
+language: c
+
+matrix:
+  include:
+    - os: linux
+      addons:
+        apt:
+          packages:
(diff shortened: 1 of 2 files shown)


mist64:
Merged. I've signed up at Travis-CI.org, but I need to head back from the admin of commanderx16 before I can enable it.

PR #87, opened 2019-09-19
Set window title when emulator is started

### LINKED ISSUE #82: Missing window title on Linux
When running the emulator on Linux there is no window title defined.  Attached is a screenshot for reference.

Screenshot_2019-09-18_17-43-27

blackknight36:
This fixes #82


blackknight36:
@mist64 I've updated the window title as requested.

PR #89, opened 2019-09-19
Use SDL_GetBasePath to determine the default ROM path

### LINKED ISSUE #63: rom files not reading from executable directory
On Windows, if the rom and char files are not specified, the emulator tries to read them from the current directory, not the executable directory (as included in argv[0] of the command line),

blackknight36:
Fixes #63.  When no rom file is specified on the command line the
emulator will search in the executable's path for the default
ROM and char ROM files.

### DIFF
--- a/main.c
+++ b/main.c
@@ -285,9 +285,9 @@ main(int argc, char **argv)
 	uint16_t trace_address = 0;
 #endif
 
-	char *rom_filename = "/rom.bin";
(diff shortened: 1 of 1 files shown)


blackknight36:
I will test this on my Mac later tonight.

mist64:
Tested on Mac, Linux, Windows.

PR #64, opened 2019-09-15
CMake and Visual Studio support

vii1:
CMake makefile.
Some conditionals for MSVC which lacks some POSIX headers.
Provided alternatives for some GCC extensions.
Also changed SDL_WINDOW_FULLSCREEN to SDL_WINDOW_FULLSCREEN_DESKTOP, which doesn't change the screen video mode and is friendlier, at least in Windows.

PR #91, opened 2019-09-19
Set window icon when the emulator is started

### LINKED ISSUE #90: Set window icon on the emulator
I've been able to get a default window icon set however I don't have the pixel data for the logo.  Does anybody have a copy of the logo as a 16x16 bitmap?  Right now I'm using the default SDL icon as shown here.

Screenshot_2019-09-19_11-46-52

blackknight36:
This window icon is based on the image displayed on the web site.  Fixes #90


mist64:
Works on Mac, but
* the file needs to be taken from the directory the executable resides in. main.c already has logic for this.
* does SDL support loading a PNG instead of a BMP?
* asterisk with the variable name, not the type, to be consistent with the existing code :P

FreeFull:
Should the icon maybe be embedded inside the binary, rather than being a separate file?

blackknight36:
@FreeFull That is possible I just thought it would be easier to store the data in a file.  I'll work on making the changes requested tomorrow.

blackknight36:
It appears that the SDL base library only support bitmaps however there is the SDL_image library which supports other image formats.

http://gigi.nullneuron.net/gigilabs/loading-images-in-sdl2-with-sdl_image for example.

@mist64 I'm not sure if including the image path matters since the data gets compiled into the binary.

madebr:
My 2c:
- Create a script that generates a header/source file from a file (python?)
- Add a build rule to the build system
- Compile and link the generated source
- Use [SDL_LoadBMP_RW](https://wiki.libsdl.org/SDL_LoadBMP_RW) to use the file.

blackknight36:
@madebr I've actually been working on a lua script to export bitmap data as hex values however I haven't quite been able to make it work yet.  It's a lot tougher dealing with binary data than I thought.  😄

madebr:
@blackknight36 #64 is adding CMake support to enable building on other platforms such as Windows and finally adding CI for all platforms.
Python is standard available on all these platforms.
I've taken a try at it and code is at this [gist](https://gist.github.com/madebr/05fdbdb7527066e972582eef656916b1).
Use it as `./embed_c.py icon.bmp -o icon.c -p icon.h`

blackknight36:
@madebr Thanks, I'll give that a shot.  I've almost got my own script working as well, just need to fix a few bugs when dealing with padded bitmap data.

video.c
```diff
@@ -183,8 +187,22 @@ video_init(uint8_t *in_chargen, int window_scale, char *quality)
 				       SDL_TEXTUREACCESS_STREAMING,
 				       SCREEN_WIDTH, SCREEN_HEIGHT);
 
+	char icon_path_data[PATH_MAX];
+
+	char *icon_filename = "icon.bmp";
+	char *icon_path = icon_path_data;
```
mist64: `char *icon_path[PATH_MAX];`

and kill line 190

video.c
```diff
@@ -183,8 +187,22 @@ video_init(uint8_t *in_chargen, int window_scale, char *quality)
...
+	char *base_path = SDL_GetBasePath();
+
+    memcpy(icon_path, base_path, strlen(base_path));
```
mist64: tabs instead of spaces, please

video.c
blackknight36: I've removed line 190 however changing line 192 to use ```char *icon_path[PATH_MAX];``` results in an incompatible pointer type error.  I've updated this PR with the other changes suggested however and I'm able to build the emulator with no errors on my Linux system.

blackknight36: I've changed icon_filename to a const char array.

mist64:
Sorry I haven't commented on this for a while. As I understand the code, it doesn't do any error checking if the file is not found, right?

blackknight36:
@mist64 That is correct.  I can add error checking if needed.

blackknight36:
@madebr I've updated this PR to use SDL_LoadBMP_RW based on the data created by the embed_c.py script.  I had to make a few slight changes to get the emulator to build but it appears to be working now.

Makefile
```diff
@@ -34,9 +34,9 @@ ifdef EMSCRIPTEN
 	OUTPUT=x16emu.html
 endif
 
-OBJS = cpu/fake6502.o memory.o disasm.o video.o ps2.o via.o loadsave.o spi.o vera_spi.o sdcard.o main.o debugger.o javascript_interface.o joystick.o
+OBJS = cpu/fake6502.o memory.o disasm.o video.o ps2.o via.o loadsave.o spi.o vera_spi.o sdcard.o main.o debugger.o javascript_interface.o joystick.o icon.o
 
-HEADERS = disasm.h cpu/fake6502.h glue.h memory.h video.h ps2.h via.h loadsave.h joystick.h
+HEADERS = disasm.h cpu/fake6502.h glue.h memory.h video.h ps2.h via.h loadsave.h joystick.h icon.h
```
madebr: Add a rule to generate `icon.c` and `icon.h` as following:
```
icon.c icon.h: icon.bmp tools/embed_c.py
	python3 tools/embed_c.py icon.bmp -o icon.c -h icon.h -n icondata
```
That way, there is no need to add icon.c and icon.h to the repo.

Add `icon.c` and `icon.h` to `.gitignore`

video.c
```diff
@@ -156,8 +161,14 @@ video_init(int window_scale, char *quality)
 									SDL_TEXTUREACCESS_STREAMING,
 									SCREEN_WIDTH, SCREEN_HEIGHT);
 
+	SDL_RWops *rw = SDL_RWFromMem(icondata, sizeof(icondata));
```
madebr: Use [SDL_RWFromConstMem](https://wiki.libsdl.org/SDL_RWFromConstMem) and the const can stay.

tools/embed_c.py
```diff
@@ -0,0 +1,63 @@
...
+    if ns.header:
+        cout.write('#include "{}"\n'.format(ns.header))
+        cout.write('\n')
+    cout.write('char {}[] = {{\n'.format(ns.name))
+    while True:
+        line = inputfile.read(8)
+        data_length += len(line)
+        if not line:
+            break
+        cout.write('    {}\n'.format(' '.join('0x{:02x},'.format(b) for b in line)))
```
madebr: Replace the four spaces with a single `\t`.

tools/embed_c.py
```diff
@@ -0,0 +1,63 @@
...
+    inputfile = ns.input
+    if hasattr(inputfile, 'buffer'):
+        inputfile = inputfile.buffer
+
+    # Write the C file
+    cout = ns.output
+    if ns.header:
+        cout.write('#include "{}"\n'.format(ns.header))
+        cout.write('\n')
+    cout.write('char {}[] = {{\n'.format(ns.name))
```
madebr: When using `SDL_RWFromConstMem`, the const can remain.

tools/embed_c.py
```diff
@@ -0,0 +1,63 @@
...
+    # Write the H file
+    if ns.header:
+        guard_trans = str.maketrans('.@,-=+', '______')
+        guard_name = '_{}_'.format(ns.header.upper().translate(guard_trans))
+
+        with open(ns.header, 'w') as hout:
+            hout.write('#ifndef {}\n'.format(guard_name))
+            hout.write('#define {}\n'.format(guard_name))
+            hout.write('\n')
+            hout.write('extern char {}[{}];\n'.format(ns.name, data_length))
```
madebr: When using `SDL_RWFromConstMem`, the const can remain.

video.c
blackknight36: I've changed this to use SDL_RWFromConstMem

tools/embed_c.py
blackknight36: Script has been redownloaded from the gist.

Makefile
blackknight36: I've updated the Makefile to generate the header files during the build.  This does introduce a build dependency on python however.

blackknight36:
I've removed the icon.xcf file which now lets the build pass.

mist64:
Thanks for your work, and sorry this is taking so long. I don't want to have a dependency on Python for the build process, so can you check in the resulting .c/.h files in addition to the script and remove the dependency in the Makefile (just make it an optional target)?

blackknight36:
@mist64 No problem, I've updated this branch as suggested.

mist64:
CI is unhappy because of spaces vs. tabs in icon.c

blackknight36:
> CI is unhappy because of spaces vs. tabs in icon.c

I've updated the script to use tabs instead of spaces.  CI checks should work now.

blackknight36:
This is no longer needed.

PR #94, opened 2019-09-20
Allow specifying image scaling quality

mporrato:
This patch enables the use of alternative SDL scaling algorithms to improve image quality when using a non integer scaling ratio (i.e.: when manually resizing the emulator window).

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,6 +29,11 @@ You can start `x16emu`/`x16emu.exe` either by double-clicking it, or from the co
 * `-run` executes the application specified through `-prg` or `-bas` using `RUN` or `SYS`, depending on the load address.
 * `-scale` scales video output to an integer multiple of 640x480
 * `-echo` causes all KERNAL/BASIC output to be printed to the host's terminal. Enable this and use the BASIC command "LIST" to convert a BASIC program to ASCII (detokenize).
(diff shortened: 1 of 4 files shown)


mist64:
I'm a little confused. Is this about the speed/quality tradeoff, or about a certain intended look?
* if it's about speed/quality: VERA emulation is  so horribly slow (dozens of instructions per pixel)  that any scaling logic (a few instructions per pixel) won't matter
* if it's about the look, then "low"/"medium"/"high" are confusing.

mporrato:
This is not intended to address the video emulation speed at all.
It's only about getting a better quality when scaling the output window  to non integer multipliers. It does not make a noticeable difference when using a HiDPI display but it makes the experience significantly better with 1080p or smaller resolutions.
The three different settings are there to allow some control over the trade-off between speed and quality of the scaling filter. The default setting (low) has exactly the same behaviour as the current code.
I could use the SDL terminology of "nearest/linear/best" if you think it makes more sense: https://wiki.libsdl.org/SDL_HINT_RENDER_SCALE_QUALITY .

mist64:
Yes, I'd prefer "nearest/linear/best",  and can we default to "best"?

sebastianvog:
I just tried that out. it looks great !

PR #95, opened 2019-09-20
fixed broken path, because trailing zero was not copied

Frank-Buss:
path

### DIFF
--- a/main.c
+++ b/main.c
@@ -307,7 +307,7 @@ main(int argc, char **argv)
 
 	// This causes the emulator to load ROM data from the executable's directory when
 	// no ROM file is specified on the command line.
-	memcpy(rom_path, base_path, strlen(base_path));
+	memcpy(rom_path, base_path, strlen(base_path) + 1);
 	strncpy(rom_path + strlen(rom_path), rom_filename, PATH_MAX - strlen(rom_path));
 #ifndef VERA_V0_8
 	strncpy(char_path, base_path, PATH_MAX);


mist64:
(Is it too late to rewrite everything in Rust?)

Frank-Buss:
Right, this would fix it, too. I tend to avoid pointer arithmetics, maybe better to create normal strings as with my patch and then simply use strcat, but which of course could fail if the path is getting too long. Rust is looking interesting :-)

mist64:
I had switched from strncpy() to memcpy() because GCC 8 was issuing a warning (-Wstringop-truncation), and creative googling didn't result in any better solution. Then again, maybe turning off the warning for this one line would have been the better solution.

mist64:
Hah!

When switching to this:
```
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Wstringop-truncation"
	strncpy(rom_path, base_path, PATH_MAX);
#pragma GCC diagnostic pop
```

LLVM will warn it doesn't know GCC's warning – and throw an error because of -Werror. :)

```
main.c:311:32: error: unknown warning group '-Wstringop-truncation', ignored [-Werror,-Wunknown-warning-option]
```

I can't use -Wno-unknown-warning-option, because that one is unsupported by GCC. :)

Sure, switching to a better build system (CMake) can take care of things like these more easily, but I'd like to keep everything as simple as possible for as long as possible...

greg-king5:
`base_path` is an existing path-name, but without the program's file-name at the end.  Therefore, it's _guaranteed to be shorter_ than `PATH_MAX`!  It doesn't need to be limited; just use `strcpy()`.

Another thing to remember is that `PATH_MAX` is a maximum `strlen()`, not a maximum `sizeof`.  Therefore, path string arrays must be defined as `[PATH_MAX + 1]`.

PR #98, opened 2019-09-21
[WITH_YM2151] Add -sound option, and related SDL and audio code upgrades

LRFLEW:
This adds a new command-line option `-sound <output device>`, which allows you to specify the output device used to play the emulated audio. This is a minor feature that became trivial to implement with some code restructuring I ended up doing.

I replaced `SDL_OpenAudio()` with `SDL_OpenAudioDevice()`. The former really only exists to make updating SDL 1.2 programs easier, and the functions with the `Device` suffix are generally intended to be used for new code. One nice advantage to this function is that you can specify exactly which parameters in the provided `SDL_AudioSpec` are required by your program, and allows SDL to perform conversions behind-the-scenes for you if the requested format isn't directly supported. This allowed me to remove the check to make sure the audio device we get is using signed 16-bit samples. I delayed the initialization of the YM2151 emulator until after getting the audio device so I could initialize it with the sample rate we actually got. I also added an explicit call to close the opened audio device to have proper cleanup.

I moved the calls to `SDL_Init()` and `SDL_Quit()` from `video.c` to `main.c`. Since SDL is starting to be used by components other than video (audio, and possibly joystick later), initializing SDL in `main()` makes sense. I also moved the call to `video_end()` from `emulator_loop()` to `main()` to match the location of `SDL_Quit()`. Lastly, along with adding `SDL_INIT_AUDIO` to `SDL_Init()` (required to use the `Device` versions of the audio functions), I added `SDL_INIT_EVENTS` since we do use that subsystem (It's implicitly initialized when initializing video, so it's not strictly required).

### DIFF
--- a/main.c
+++ b/main.c
@@ -345,7 +345,7 @@ void initAudio()
 	}
 
 	// init YM2151 emulation. 4 MHz clock
-	YM_Create(1.0f, 4000000);
+	YM_Create(4000000);
 	YM_init(have.freq, 60);
 
 	// start playback
--- a/main.c
+++ b/main.c
@@ -92,6 +92,11 @@ FILE *prg_file ;
 int prg_override_start = -1;
 bool run_after_load = false;
 
+#ifdef WITH_YM2151
+const char *audio_dev_name = NULL;
+SDL_AudioDeviceID audio_dev;
+#endif
+
 #ifdef TRACE
 #include "rom_labels.h"
 char *
@@ -273,7 +278,10 @@ usage()
 	printf("-dump {C|R|B|V}...\n");
 	printf("\tConfigure system dump: (C)PU, (R)AM, (B)anked-RAM, (V)RAM\n");
 	printf("\tMultiple characters are possible, e.g. -dump CV ; Default: RB\n");
-
+#ifdef WITH_YM2151
+	printf("-sound <output device>\n");
+	printf("\tSet the output device used for audio emulation");
+#endif
 #ifdef TRACE
 	printf("-trace [<address>]\n");
 	printf("\tPrint instruction trace. Optionally, a trigger address\n");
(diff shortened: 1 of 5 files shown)


LRFLEW:
Went ahead and removed the volume setting from the YM2151 emulator, as done in #105. Looking at it, I agree with @Frank-Buss that the additional int->float->int conversion required is not worth it. Additionally, a proper volume control should probably do some sort of dithering. I don't have a setup yet for WASM testing, so I don't know if the other part of that PR is still necessary.

LRFLEW:
One last quick change. I replaced `AUDIO_S16` with `AUDIO_S16SYS`. For 99% of systems, this will be exactly the same. The difference is that `AUDIO_S16SYS` specifies "system endianness", which is what is actually being used.

Frank-Buss:
Good idea. This will allow it to run properly on SPARC and 68000 systems as well.

main.c
```diff
@@ -254,30 +262,39 @@ void initAudio()
 	SDL_AudioSpec want;
 	SDL_AudioSpec have;
 
-	// init YM2151 emulation. 4 MHz clock
-	YM_Create(1.0f, 4000000);
-	YM_init(SAMPLERATE, 60);
+	/*
```
mist64: can we print the list if -sound is passed without an argument? -keymap does something similar.

mist64:
I like it, apart from a minor comment. (Also, there are conflicts now, sorry about that.)

LRFLEW:
Rebased to fix the conflicts, and now have the `-sound` argument print the sound devices when no argument is passed for it or the sound device isn't found.

LRFLEW:
Forgot to add an entry to the README for the new option. Fixed now.

PR #100, opened 2019-09-21
[WebAssembly] Better browser support

sebastianvog:
Webkit, Safari, Chrome on Android and Opera supported now.
Removed threading as it is not supported currently on browsers other than Chrome Desktop.
Changed Main emulator loop to yield control back to the browser after a complete frame.
Setting emscripten main loop fps parameter to 0, as this is highly recommended in emscripten documentation.
Lastly, reworked imports for *sdcard_file, due to a compile error using the llvm webassembly backend.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,7 +28,7 @@ endif
 
 ifdef EMSCRIPTEN
 	# Todo #--js-library webassembly/helper.js
-	LDFLAGS+=--shell-file webassembly/x16emu-template.html  --preload-file rom.bin --preload-file chargen.bin -s DISABLE_DEPRECATED_FIND_EVENT_TARGET_BEHAVIOR=1 -s USE_PTHREADS=1 
+	LDFLAGS+=--shell-file webassembly/x16emu-template.html  --preload-file rom.bin -s DISABLE_DEPRECATED_FIND_EVENT_TARGET_BEHAVIOR=1 
 	# To the Javascript runtime exported functions
 	LDFLAGS+=-s EXPORTED_FUNCTIONS='["_j2c_reset", "_j2c_paste", _main]' -s EXTRA_EXPORTED_RUNTIME_METHODS='["ccall", "cwrap"]'
 	
--- a/sdcard.c
+++ b/sdcard.c
@@ -6,6 +6,7 @@
 #include "sdcard.h"
 
 static int cmd_receive_counter;
+FILE *sdcard_file = NULL;
 
 void
 sdcard_select()
--- a/main.c
+++ b/main.c
@@ -43,7 +43,7 @@
 
 
 void* emulator_loop(void *param);
-void sdl_render_callback(void);
+void emscripten_main_loop(void);
 
 
 // This must match the KERNAL's set!
@@ -559,9 +559,7 @@ main(int argc, char **argv)
 	instruction_counter = 0;
 
 #ifdef __EMSCRIPTEN__
-	pthread_t tid;
-    pthread_create(&tid, NULL, emulator_loop, NULL);
-	emscripten_set_main_loop(sdl_render_callback, 60, 1);
+	emscripten_set_main_loop(emscripten_main_loop, 0, 1);
 #else
 	emulator_loop(NULL);
 #endif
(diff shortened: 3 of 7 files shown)


sdcard.c
```diff
@@ -6,6 +6,7 @@
 #include "sdcard.h"
 
 static int cmd_receive_counter;
+FILE *sdcard_file = NULL;
```
sebastianvog: If someone could take a look at the sdcard_file declarations/imports ... To be honest, I am slightly unsure about the importing (I really should refresh my C knowledge). Reason I am doing it that i got "asm-ld: error: duplicate symbol: sdcard_file" error when compiling with latest WebAssembly llvm.

Makefile
```diff
@@ -28,7 +28,7 @@ endif
 
 ifdef EMSCRIPTEN
 	# Todo #--js-library webassembly/helper.js
-	LDFLAGS+=--shell-file webassembly/x16emu-template.html  --preload-file rom.bin --preload-file chargen.bin -s DISABLE_DEPRECATED_FIND_EVENT_TARGET_BEHAVIOR=1 -s USE_PTHREADS=1 
+	LDFLAGS+=--shell-file webassembly/x16emu-template.html  --preload-file rom.bin --preload-file chargen.bin -s DISABLE_DEPRECATED_FIND_EVENT_TARGET_BEHAVIOR=1
```
madebr: The `chargen.bin` file is now included in `rom.bin`: https://github.com/commanderx16/x16-emulator/pull/64#issuecomment-533674910

madebr:
I have tested this on firefox using cmake and emscripten 1.38.29 (I used conan package manager).
My build has a big lag and audio is not working.

Makefile
sebastianvog: Will address , thank you

sebastianvog:
> I have tested this on firefox using cmake and emscripten 1.38.29 (I used conan package manager).
> My build has a big lag and audio is not working.

thank you for testing this. 
Are you talking about compile time or execution speed in the browser? The browser is unfortunately lacking quite behind in speed. I started looking into optimizations but so far could only find a few percentage % improvements by using the emscripten llvm backend. I think i would have to profile to get a better understanding where the time is spent.
I have not looked at audio yet, I can check if that would be something easy to address. or could be limitation by emscripten/sdl implementation, not sure.

madebr:
I was talking about input lag but nevermind, it is working fine now. There must have been a background process going on.
Sorry for the noise.

madebr:
I don't know how, but the emulator at http://www.frank-buss.de/x16/x16emu.html has working copy/paste on firefox.

sebastianvog:
> http://www.frank-buss.de/x16/x16emu.html

@Frank-Buss modified getClipboard() function to read the text from the textarea and pass it to the emulator. the original code reads from navigator.clipboard.readText() and that call unfortunately doesn't work on Firefox.

Frank-Buss:
I think an additional textarea would be nice. Not only as a workaround for the clipboard functionality, but if the focus works properly, someone could even develop longer BASIC programs in the textarea and when done, can copy it elsewhere to save it. I guess currently there is no other way to get something out of the emulator, and it is lost when the browser is closed?

Tmp2k:
I'm working on a simple BASIC web IDE at the moment, the concept being the page is split 50/50 with a basic editor on the left and the emulator on the right.  Rather than just pasting in text, you can use the editor and hit "Run".  You could add on the ability to save to local disk, or whatever you like using JS.

If people would find this useful I can add the basics into this repo, rather than starting my own?

My plan is to include some compilers in the long run, allowing you to code in C, ASM or BASIC and have it run instantly on-page using the emulator. (if anyone wants to help with this, I'm just trying to get to grips with the emscr virtual file system)

sdcard.c
madebr: Your patch is correct. `extern`ed variables should go in the headers.

mist64:
Excellent, thanks!

Tmp2k:
BTW, I've fixed the focus / keypress issue and added an extra textinput for entering basic. I've also got it to auto detect the key map from the browser locale. Plus a few other tweaks. I started with the code from this pull request as it already fixed a lot of issues. I'll submit a new pull request this evening.

sebastianvog:
That’s excellent! I couldn’t figure out the keyboard/ focus issue and detectIng the keyboard Layout is great !

PR #105, opened 2019-09-21
removed not needed volume setting and adding float support for audio

Frank-Buss:
@sebastianvog with this patch, it should work for the wasm target. At least the SDL init functions don't fail anymore, and I tested the float functionality for the normal target and it worked. But it didn't work in my browser (Chrome on Linux). But this example worked:
https://googlechromelabs.github.io/web-audio-samples/audio-worklet/design-pattern/wasm/
Maybe you have an idea.
You can compile it with audio support, if you set the environment variable `WITH_YM2151` to `1` before calling `make wasm`. You can test it with this program:
http://www.frank-buss.de/x16/audio.prg

### DIFF
--- a/main.c
+++ b/main.c
@@ -87,6 +87,10 @@ FILE *prg_file ;
 int prg_override_start = -1;
 bool run_after_load = false;
 
+#ifdef WITH_YM2151
+int audio_float = 0;
+#endif
+
 #ifdef TRACE
 #include "rom_labels.h"
 char *
@@ -246,7 +250,9 @@ usage_keymap()
 #ifdef WITH_YM2151
 void audioCallback(void* userdata, Uint8 *stream, int len)
 {
-	YM_stream_update((uint16_t*) stream, len / 4);
+	len /= 4;
+	if (audio_float) len /= 2;
+	YM_stream_update(stream, len, audio_float);
 }
 
 void initAudio()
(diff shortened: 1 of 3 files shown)


LRFLEW:
This will likely result in conflicts with my other audio-related PR #98. My PR *should* fix the WASM support by having SDL doing a back-end conversion from S16 to Float, but adding native support is probably a good idea.

Frank-Buss:
I think doing the backend conversion is better, because less clutter for the application itself. It shouldn't need more time, only a little bit for an additional small buffer, because I'm doing a conversion from S16 to float anyway, so my code has no benefit, and currently it doesn't work anyway for WASM.
I would suggest the following: @LRFLEW adds the part of my patch regarding the remove of the volume function, because this adds an unnecessary floating point multiplication, and adds this to his/her patch. I'll close my pull-request.

Frank-Buss:
PS: I just tried @LRFLEW's patch, it doesn't fix the WASM problem, but still a better patch.

sebastianvog:
I will look at the Wasm issue from a JavaScript/browser perspective .  I have seen a JavaScript warning about audio context permission.  Chrome doesn’t want audio to start without a click.  Not sure if that is the issue but perhaps it brings us a step closer

Issue #51, opened 2019-09-14
Linux CPU 100% usage


mlongval:
Great work gentlemen. 
I was testing the latest r30 build on Linux. Works great but it grabs one cpu core and runs it at 100%, even when the emulator is idle. Otherwise lots of fun.

indigodarkwolf:
Not seeing this on Windows with an i7-9700K @ 3.6GHz. The emulator seems to favor a single core, and I can eventually get it up to 100% usage if I enable a bunch of local bells and whistles and debugging display stuff that I modified into a local build, but if I just run the official r30 release I waffle between 50%-60% of one core. I don't foresee much opportunity to parallelize anything, but maybe some profiling would reveal some inefficiencies to address.

bsutherland:
I'm seeing high (but not quite 100%) usage of a single core on Windows as soon as I launch the application. i7-4710 @ 2.50 GHz. Release r30.

mobluse:
I have the same problem in Raspbian Buster on Raspberry Pi 4 B (4 cores) with 4 GB RAM (of which 256 MB is GPU memory). I see it in `htop`, but the core that is 100% switches during run ─ it can be 1, 2, 3, or 4. `x16emu` is just idle. The C64 emulator `x64` from VICE 3.3 runs at about 88% in total, but maximum 92% on one core sometimes.

indigodarkwolf:
@mobluse Maybe. That seems seems worth investigating. 

I did take a brief crack and reducing the amount of math that was needlessly recomputed in get_pixel() (because honestly there's a fair bit) and also got some wins, but invalidated rectangles would probably be a great macro-level win. The tricky part is probably knowing once it's safe to mark a given rectangle as "valid", since the emulator is, by design, allowing mid-frame, even mid-scanline updates to Vera data. So any write to the Vera at any point is potentially invalidating a portion of the screen until the renderer has had a chance to cycle around.

mist64:
First of all, there is no such thing as "idle" for the X16 emulator. Unless the screen is completely turned off, it always does the same amount of work.

And I agree with @indigodarkwolf, it's the VERA emulation that is expensive, and it's far from optimizes. It's a quite straightforward implementation, and it does lots of redundant work.

It should work more like real hardware, in that it counts up addresses as opposed to calculating the current address based on X and Y.

blackknight36:
I'm noticing about 20% CPU use on Windows as well.
[...]

PR #106, opened 2019-09-21
issue #51: Draw optimizations

indigodarkwolf:
Some simple optimizations to get_pixel and get_sprite which reduce overhead from needlessly recalculating values that only change with layer and sprite attribute changes. Current layer and sprite properties are calculated into structs that can be referenced instead of performing the math for each pixel.

In my profiling, the largest single hotspot remaining in get_pixel() and get_sprite() seems to be the calls to video_space_read(), and my best guess is that branch instructions are causing needless stalling when it would be faster to perform a bitwise-and in the special cases of get_pixel() and get_sprite(). That last tweak is pending a question in issue #51.


hasenbanck:
The commit history of this PR looks like you used PRs on your own repository to sync the current branch into your feature branch. I think it would be best if you squash your commits and rebase them against the current master and do a force push. That way we keep the commit history clean.

indigodarkwolf:
Assume I don't know what I'm doing in git. Do you mean I should discard my fork, re-fork, and essentially paste the changed file into the clean fork?

mist64:
If it’s just a single file, pasting the file into a clean fork would be easiest. :-) (git is hard!)

indigodarkwolf:
Git _is_ hard. >.> And my dayjob uses Perforce.

PR #113, opened 2019-09-22
Updated WASM demo, added lots of features, tidied up a little

Tmp2k:
I've updated the WASM demo so it's now a side-by-side editor and emulator. New features include..

* Autodetect keymap from browser locale
* Fixed canvas trapping all keystrokes/stealing focus (comments about this are in the code) 
* You can now edit BASIC in the browser and push to the emulator
* Added fullscreen mode (only works on a landscape screen for now)
* Updated WASM readme with a bit more info
* Separated out JS from HTML, generally neatened things up a bit and added some comments
* Status panel is a bit more verbose and also includes JS console messages


Tmp2k:
There's a live demo here if you want to try before you pull ;) https://x16wide.nemiah.uk

Tmp2k:
There's a live demo here if you want to try before you pull ;) https://x16wide.tmp2k.com

mist64:
Looks great, but @sebastianvog should give his okay!

sebastianvog:
Great work! Looks so much nicer!
I tested on
Desktop Mac: Chrome, Firefox and Opera.
Mobile Phone: Safari Iphone

Here two issues I encountered:
On Safari on Desktop I noticed the textarea for the code does not size correctly vertically, only two lines of the texture show . it looks like safari doesn't support the flex-grow properly?

On Safari mobile (iPhone) the emulator, it does not seem to work anymore:
Console shows
"Assertion failed: the Module object should not be replaced during async compilation- perhaps the order of HTML elements is wrong?? ( this going to https://x16wide.tmp2k.com/ )

Tmp2k:
Unfortunately, I don't have an iPhone or Mac at home, but I'll be in the office tomorrow and I'll have a look at it then. I'm in the UK so it's 00:10 now ;)

sebastianvog:
When you read this: Happy Monday :-)

After some googling/trial and error,  I  came up with adding display:flex to .editor-middle
and removing height from # code

.editor-middle {
    flex-grow: 100;
   display: flex;
}
and removing height from #98 

#code {
    /* height: 100%; */
    width: 100%;
    padding: 0;
    margin: 0;
    border: 0;
    background: #222;
    color: #999;
    resize: none;
    font-size: 16px;
    opacity: 0.8;
}

This seems to work on Chrome and Safari.  Let me know what you think.

I will try to dig into the other issue

mist64:
@sebastianvog Do you want me to merge this, and you can send me a pull request with updates?

Tmp2k:
So the assertion failed issue will be because the code that emscripten injects into the template uses the async flag but there's no callback.  So it doesn't matter what order you call the scripts, they load in parallel, I'll look into it when I get some time. We either need to detect when the emscr JS has loaded, or change how we interact with the Module object.

sebastianvog:
you would think it should work if our script loads synchronously first with 
<script src="webassembly/main.js"></script>
and then after that the x16emu.js could load asynchronously  and start doing its wasm thing

I wonder if iOS browser has some kind of optimization and loads webassembly/main.js asynchronously

I can reproduce the issue on desktop Safari,  if i load main.js after the async x16emu.js statement.

Anyways, I came down with a cold and am not thinking clear.

sebastianvog:
Changing main.js to
 var Module = {
instead of just 
Module = {

Fixes the issue of "Assertion failed: the Module object should not be replaced during async compilation... "

@Tmp2k could you make those two changes and then we would be good to go!
@mist64 alternatively we could also add those in a follow up commit. Either way works for me.

Thank you. I am excited about this PR!

Tmp2k:
I can do that later tonight if you bear with me, or merge this PR and I'll send another later. I'll probably make it look a bit nicer on portrait mobile too.

Tmp2k:
Done, I'll add some mobile styles later on. @mist64 if you merge this, I can do a separate PR for the mobile styles later.

Tmp2k:
I've updated my fork with some mobile device layout tweaks but when I try and create another pull request it's only letting me request all 6 commits from this PR and the new commit.

What am I missing?

sebastianvog:
Probably because your commits were squashed into one commit 54d0496. I would try rebasing to master or create a new branch from master and put your new commit on top.

PR #135, opened 2019-09-27
Adds a new -echo option: -echoraw

mobluse:
Option -echoraw is needed for some filter programs. It doesn't change CR to LF, or anything else.

Example of filter that needs this feature in order to make the output in the terminal look the same as on the screen: https://github.com/mobluse/x16-petscii2utf8

This feature would also be needed if one wants to directly output all types of binary files from the emulator.

### DIFF
--- a/main.c
+++ b/main.c
@@ -70,6 +70,7 @@ bool dump_ram = true;
 bool dump_bank = true;
 bool dump_vram = false;
 bool echo_mode = false;
+bool echo_raw = false;
 bool save_on_exit = true;
 bool record_gif = false;
 char *gif_path = NULL;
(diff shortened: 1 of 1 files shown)


mist64:
I changed this a bit on top of your work. It's now `-echo`, `-echo raw` and `-echo iso`. `-echo` also escapes non-printable characters (and pasting will un-escape them), and `-echo iso` is the equivalent for ISO-8859-1 plus UTF-8 conversion.

All this is necessary to support https://github.com/commanderx16/x16-emulator/issues/133 (conversion between BAS and PRG for ASCII and ISO).

I still think the same for PETSCII should be an independent script though.

Issue #56, opened 2019-09-14
Request: Hook up SDL's Joystick interface to the x16 joystick interface


ElfEars:
I think testing joysticks would be much easier if we could use IRL joysticks, especially for two player. (Plus it would be a good feature in general, your arms get tired after a long day of typing!)

Obviously there would need to be some config file/command line magic to setup what type we're emulating (NES, Keyboard Cursors, SNES) and mapping IRL joystick buttons to virtual joystick buttons but otherwise SDL already comes with good cross platform joystick support.

mist64:
For a start, I suggest the following command line options:
`-joy1 nes`
`-joy1 snes`
`-joy1 none`
`-joy2 nes`
`-joy2 snes`
`-joy2 none`
`none` would be the default for both. For now, we should just use a standard (non-configurable) mapping.

We should not map an SDL joystick to emulated keyboard keys, that would be  just  too confusing. Cursor keys and Ctrl, Alt, Space and Return are already detected by the system ROM as a fallback joystick 1 on both the emulator and on real hardware.

mobluse:
I succeeded with connecting a cheap noname SNES USB gamepad to X16 in Raspbian Buster. First I had to name the window:
`xdotool search --onlyvisible --pid $(ps aux | grep x16emu | cut -c 9-15 | head -1) --name "\a\b\c" set_window --name 'X16 Emulator'`

Then I use the SNES as a NES keyboard joystick. The buttons that are not used send their button number in Linux (which starts at 0). The joystick has only 8 physical buttons, but sends 10.
`joy2key 'X16 Emulator' -X -dev /dev/input/js0 -buttons 0 1 Alt_L Control_L 4 5 6 7 space Return -axis Left Right Up Down -thresh -9 9 -9 9`

I cannot guarantee I got the A and B buttons right because I never had SNES och NES. SNES B = NES A, and SNES Y = NES B, if I understand correctly.

In Linux one can remap the buttons in the OS using jstest-gtk, or just jstest on the command line.

jjbliss:
I'm looking into getting SDL Joysticks working on the emulator.  I've got some code writing to VIA that the kernal is picking up as joystick inputs.  I'll try to have something mostly working by tomorrow.

Does it matter for anything else where in the emulator loop I call into the controller code?  I haven't seen any problems yet, but there might a code order issue that I'm not aware of.

tobier:
Shouldn't this be closed if a PR was merged?

PR #138, opened 2019-09-27
Add SDL Controller Support

jjbliss:
As requested in issue #56 , here is SDL controller support.
To use, add a command line argument, like "-joy1 SNES" and it will use the first SDL controller.
This works in a similar way to the ps2 implementation.

This might be improved by switching over to SDL events to maintain the controller state, as I've seen it miss very fast button presses.  That issue was with a somewhat laggy bluetooth controller, so I'm not sure how much was just the bluetooth problems though.


jjbliss:
I've been testing this on MacOS, so it's possible there will be problems on other platforms.  I will go ahead and try it on Linux now.

mist64:
Great work! Just minor comments:
* Can you please be careful with the coding style. Tabs like everywhere else, spaces before and after "{" and "}", space after "if", "i++" instead of "++i". If in doubt, copy the coding style in the rest of the project.
* We don't need "-joy1 none", since this would be the default without specifying "-joy1".
* We need to be consistent about whether we name it "controller" or "joystick". In a Commodore world, it's a joystick, and that's what the KERNAL calls it. I think we should stick with it, so the C symbols, filenames and the help text should say joystick.

joystick.c
```diff
@@ -0,0 +1,178 @@
...
+		}
+	}
+	writing = false;
+	return true;
+}
+
+void joystick_step()
+{
+	if (joy1_mode == NONE) {
+		return;
```
greg-king5: That code doesn't let us use _only_ `-joy2`!  (We might want the keyboard to be JOY1.)

joystick.c
```diff
@@ -0,0 +1,178 @@
...
+	}
+	if (!writing) { //if we are not already writing, check latch to
+		//see if we need to start
+		handle_latch(joystick_latch, joystick_clock);
+	}
+
+	//if we have started writing controller data and the latch has dropped,
+	// we need to start the next bit
+	if (writing & !joystick_latch) {
```
greg-king5: The `writing` variable is tested immediately above here; it doesn't need to be tested again.  Therefore, the line would be better as
```c
else if (!joystick_latch) {
```

jjbliss:
Thanks greg, I'll put in those changes.  A few more too.
I've put together a small test program to see the controller inputs.  There are a few problems I'm seeing, so I'll try to fix those.
https://github.com/jjbliss/x16-controller-test

joystick.c
```diff
@@ -0,0 +1,178 @@
...
+static uint16_t joystick2_state = 0;
+static uint8_t clock_count = 0;
+
+
+bool joystick_latch, joystick_clock;
+bool joystick1_data, joystick2_data;
+
+bool joystick_init()
+{
+	int joystick1_number = 0;
```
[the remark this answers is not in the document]
jjbliss: I've updated this.

joystick.c
jjbliss: I've done some updates to make sure that joy2 can work at the same time as the keyboard now.  It turns out I wasn't even properly checking joy2_mode.

joystick.c
jjbliss: I have updated this.

PR #150, opened 2019-09-30
New command line switch to make it easier to load local programs

izb:
Add -mountdir CLI switch which changes the host directory after the ROM has loaded. Useful for when you have multiple versions of the emulator, but need to load programs or other files from a common location.

### DIFF
--- a/main.c
+++ b/main.c
@@ -283,7 +283,7 @@ usage()
 	printf("\tEnable a specific keyboard layout decode table.\n");
 	printf("-sdcard <sdcard.img>\n");
 	printf("\tSpecify SD card image (partition map + FAT32)\n");
-	printf("-mountdir <directory>\n");
+	printf("-dir <directory>\n");
 	printf("\tSpecify a directory for the host filesystem bridge (device 1).\n");
 	printf("-prg <app.prg>[,<load_addr>]\n");
 	printf("\tLoad application from the local disk into RAM\n");
@@ -510,7 +510,7 @@ main(int argc, char **argv)
 			sdcard_path = argv[0];
 			argc--;
 			argv++;
-		} else if (!strcmp(argv[0], "-mountdir")) {
+		} else if (!strcmp(argv[0], "-dir")) {
 			argc--;
 			argv++;
 			if (!argc || argv[0][0] == '-') {
--- a/main.c
+++ b/main.c
@@ -284,7 +284,7 @@ usage()
 	printf("-sdcard <sdcard.img>\n");
 	printf("\tSpecify SD card image (partition map + FAT32)\n");
 	printf("-mountdir <directory>\n");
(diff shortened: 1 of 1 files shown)


main.c
```diff
@@ -707,6 +720,16 @@ main(int argc, char **argv)
 		}
 	}
 
+	if (mountdir_path) {
+		DIR* dir = opendir(mountdir_path);
```
mist64: space, then asterisk.

main.c
```diff
@@ -282,6 +283,8 @@ usage()
 	printf("\tEnable a specific keyboard layout decode table.\n");
 	printf("-sdcard <sdcard.img>\n");
 	printf("\tSpecify SD card image (partition map + FAT32)\n");
+	printf("-mountdir <directory>\n");
+	printf("\tSpecify a directory to switch to once loaded.\n");
```
mist64: "Specify a directory for the host filesystem bridge (device 1)."

mist64:
* Good feature, thanks! Can we call it just `-dir` though?
* Two small comments in the diff.
* Also, CI failed because the compiler on Linux (GCC) insists that you check the return value of `chdir()`. You're already checking whether it exists, do just say `(void)chdir(mountdir_path);`.

greg-king5:
That patch isn't needed.  We already have that feature.  We just do it the other way:

1. Change to the Directory where we have our programs.
2. Use a path-name to choose which emulator that we want to execute.

izb:
> That patch isn't needed. We already have that feature. We just do it the other way:
> 
> 1. Change to the Directory where we have our programs.
> 2. Use a path-name to choose which emulator that we want to execute.

The emulator looks in its cwd to find the rom file, so this patch allows you to have your programs in a different location to the rom.

greg-king5:
> > That patch isn't needed. We already have that feature. We just do it the other way:
> > 
> > 1. Change to the Directory where we have our programs.
> > 2. Use a path-name to choose which emulator that we want to execute.
> 
> The emulator looks in its cwd to find the rom file, so this patch allows you to have your programs in a different location to the rom.

[Not anymore -- it was fixed!](https://github.com/commanderx16/x16-emulator/releases/tag/r31)  The emulator now looks in the directory from which its binary executable was loaded.

I use the feature all the time.

izb:
> > > That patch isn't needed. We already have that feature. We just do it the other way:
> > > 
> > > 1. Change to the Directory where we have our programs.
> > > 2. Use a path-name to choose which emulator that we want to execute.
> > 
> > 
> > The emulator looks in its cwd to find the rom file, so this patch allows you to have your programs in a different location to the rom.
> 
> [Not anymore -- it was fixed!](https://github.com/commanderx16/x16-emulator/releases/tag/r31) The emulator now looks in the directory from which its binary executable was loaded.
> 
> I use the feature all the time.

So it was! I'll close this PR

PR #157, opened 2019-10-02
Allow emulator to start/stop GIF recording.

rsbohn:
`POKE $9FB5,1` to snap one frame to the .gif file.
`POKE $9FB5,3` to start recording.
`POKE $9FB5,0` to stop recording.

Breaking Change: You have to poke to start the recording, it won't start unless you do the poke.

Why: I do lots of recordings and sometimes I don't want to include the X16 splash screen. Sometimes I want just one frame, or I want to take one snap each time my drawing loop is done.
The attached recording shows starting the recording after modifying the display (still the splash screen, but at 80x30 resolution.)
recording

### DIFF
--- a/README.md
+++ b/README.md
@@ -195,6 +195,10 @@ All rights reserved. License: 2-clause BSD
 
 ## Release Notes
 
+### Next Release
+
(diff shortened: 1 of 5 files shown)


rsbohn:
Maybe we could use `-gif filename,wait` when you want to control when the recording starts, otherwise it starts immediately. Then it wouldn't be a breaking change.

mist64:
I agree with rsbohn. The default should be to start immediately.

rsbohn:
I'll add that to main.c (based on the recent -echo raw code).

PR #158, opened 2019-10-02
Some minor enhancements on Debugger

kktos:
Here is the list of changes:
- ~~use TTF for debugger rendering~~
- displays 24bits addresses on Code and Data when in Banked Memory
- displays the virtual Bank registers (RAM/ROM)
- displays the VERA Registers (ADDR / DATA0 / DATA1 / CTRL)
- displays the top 20 bytes from stack
- adds a command line (*)
- adds PAGEUP/PAGEDOWN to browse Memory dump
- adds some DEBUG function replacement to allow to read memory without any side effects (i.e: read VERA w/o increment address)

**(*) list of commands**
- m address  
displays memory dump from this address.  
e.g:  
    m 5000 to display main memory from $5000  
    m 5A000 to display RAM memory from bank 5, address $A000  
    m 7F500 to display ROM memory from bank 7, address $F500  
- d address  
displays code disasm from this address.  
e.g:  
    d 5000  
    d 5A000  
    d 7F500  
- b rom|ram bank_number  
set the ROM bank number or the RAM bank number.  
e.g:  
b rom 7 to switch to ROM bank 7
- r A|X|Y|PC|SP number  
set value for specific register.  
r a 54 to set register Acc to value $54


memory.c
```diff
@@ -91,6 +91,67 @@ read6502(uint16_t address)
 	}
 }
 
+/*
+	Read Memory only for debugger; no I/O, no side effects
+	duplicate read6502 to avoid impacting normal use performances
+*/
+uint8_t
+DEBUGread6502(uint16_t address, uint8_t bank)
```
mist64: instead of duplicating the existing function, can you create a generic internal function that takes a "bool debug" argument and special cases only the video case, and two frontends, the regular one, and the debug one?

And the same below for the video version.

mist64:
Thanks a lot for this! I'm not a fan of TTF though. x16emu has very few dependencies so far, and this would a huge dependency. I want it be easy to build and easy to port. We run on MorphOS for example, which is awesome, and I don't want to break these use cases. I don't see anything wrong with using the pixel font. Pixel fonts are great! :)

kktos:
ok, I'll rollback the pixel font ;)

memory.c
kktos: I did it on purpose ;) Here, I thought it would be better to have a specific function for the debugger so when it's not use, the normal flow is totally unaltered vs a bool test on every call. I didn't want to have any impact on the emulator perf. That was the intention.
Now, I agree, it could be a pita to have to maintain 2 functions. Your call, mate.

paulscottrobson:
The display is much improved. I agree about the TTF fonts. The command line thing I don't care for tbh. This is a personal choice though, as I've been using this debugger style in a framework for about 10 years (I still find the keyboard assignment changes irritating, I don't normally use those function keys). Michael's call on this one. Maybe have the old version as a build option ?

kktos:
@paulscottrobson  For the display, Michael's arguments were valid therefore I rollback the pixel font.
As for the shortcuts, I've in mind since a while to make them configurable, from an external conf file, for instance. We are creatures of habits and I would be the first customer of such a configuration ;)
I've to finish the http remote controller first ;)

kktos:
btw, I'm seeing Travis is unhappy but I've no idea how to fix it...

tomseago:
@kktos if you click on the "Details" link for the Travis failure you get a page showing 3 build jobs (one per platform). Click on any one of those and you get the full build log.

In this case the issue is an implicit declaration of stricmp - so I'll guess it's just a missing header file in debugger.c?

kktos:
@tomseago Tom, you're a genius ;) Thx a lot mate, I didn't click far enough the first time I checked. my bad.

memory.c
kktos: @mist64 So Michael, what do you want me to do ?

mist64: Branch prediction will limit the cost of the bool check to practically zero. :-)

kktos: done.

PR #161, opened 2019-10-03
Allow control of .gif recording from within the emulator.

rsbohn:
-gif capture.gif (works as always)
-gif capture.gif,wait (start with recording paused)

`POKE $9FB5,1` to record a single frame.
`POKE $9FB5,3` to start or resume recording.
`POKE $9FB5,0` to stop recording.

`PEEK($9FB5)` to check the recording status. Returns 128=paused, 3=recording, 1=snapshot pending, 0=recording inactive (-gif not specified on command line, or write to .gif failed), 255=feature not present (running on actual hardware??).

In this clip the recording doesn't start until the second time through the loop.
spin

### DIFF
--- a/glue.h
+++ b/glue.h
@@ -41,7 +41,7 @@ extern bool log_video;
 extern bool log_keyboard;
 echo_mode_t echo_mode;
 extern bool save_on_exit;
-extern bool record_gif;
+extern uint8_t record_gif;
 extern char *gif_path;
 extern uint8_t keymap;
 
--- a/glue.h
+++ b/glue.h
@@ -29,6 +29,18 @@ typedef enum {
 	ECHO_MODE_ISO,
 } echo_mode_t;
 
+// GIF recorder commands
+#define RECORD_GIF_PAUSE	0
+#define RECORD_GIF_SNAP		1
+#define RECORD_GIF_RESUME	2
+
+// GIF recorder states
+// [paused] [0] [0] [0] [0] [0] [continuous] [active]
+#define RECORD_GIF_PAUSED	128
+#define RECORD_GIF_ACTIVE	3
+#define RECORD_GIF_SINGLE	1
+#define RECORD_GIF_DISABLED	0
+
 extern uint8_t a, x, y, sp, status;
 extern uint16_t pc;
 extern uint8_t *RAM;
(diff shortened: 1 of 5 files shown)


video.c
```diff
@@ -158,8 +158,19 @@ video_init(int window_scale, char *quality)
...
 	SDL_SetWindowTitle(window, "Commander X16");
 
-	if (record_gif) {
-		record_gif = GifBegin(&gif_writer, gif_path, SCREEN_WIDTH, SCREEN_HEIGHT, 1, 8, false);
+	if (record_gif > 0) {
+		if (!strcmp(gif_path+strlen(gif_path)-5, ",wait")) {
+			// wait for POKE
+			record_gif = 128;
+			// move the string terminator to remove the ",wait"
+			gif_path[strlen(gif_path)-5] = 0;
```
rsbohn: Is this a good way to trim the path?

memory.c
```diff
@@ -175,6 +175,26 @@ memory_get_rom_bank()
 	return rom_bank;
 }
 
+// Control the GIF recorder
+//        bit:  7     6  5  4  3  2  1           0
+// record_gif: [ctrl][0][0][0][0][0][continuous][active]
+void
+emu_recorder_set(uint8_t newstate)
+{
```
sebastianvog: imho it would be easier to read if the possible values(or bit mask) for record_gif would be enums or #defines instead of 0, 1, 128,255.

rsbohn: Added #defines, it does make the code clearer.

glue.h
```diff
@@ -29,6 +29,18 @@ typedef enum {
 	ECHO_MODE_ISO,
 } echo_mode_t;
 
+// GIF recorder commands
+#define RECORD_GIF_PAUSE	0
```
mist64: enum

glue.h
```diff
@@ -41,7 +53,7 @@ extern bool log_video;
 extern bool log_keyboard;
 echo_mode_t echo_mode;
 extern bool save_on_exit;
-extern bool record_gif;
+extern uint8_t record_gif;
```
mist64: enum type :)

video.c
mist64: Yup.

PR #165, opened 2019-10-06
Make load routine wrap into next RAM bank

jjgignac:
This changes the LOAD command to wrap into the next RAM bank whenever it reaches the end of the high RAM area.  This makes it easy to load files larger than 8K into high memory.

Corresponding KERNAL change: https://github.com/commanderx16/x16-rom/pull/53.

Credit for the idea goes to [Artur Bujdoso](https://www.facebook.com/groups/CommanderX16/permalink/517307852353647/?comment_id=517330952351337) and/or [Tom Wilson](https://www.facebook.com/groups/CommanderX16/permalink/524303641654068/?comment_id=524556388295460).


loadsave.c
```diff
@@ -169,7 +169,15 @@ LOAD()
...
 			// banked RAM
-			bytes_read = fread(RAM + ((uint16_t)memory_get_ram_bank() << 13) + start, 1, 0xc000 - start, f);
+			while(1) {
+				size_t len = 0xc000 - start;
+				bytes_read = fread(RAM + ((uint16_t)memory_get_ram_bank() << 13) + start, 1, len, f);
+				if(bytes_read < len) break;
+
+				// Wrap into the next bank
+				start = 0xa000;
+				write6502(0x9f61, 1 + read6502(0x9f61));
```
mist64: Better use `memory_set_ram_bank()`/`memory_get_ram_bank()`, which are already exported.

jjgignac: The problem is that those functions don't update `via1registers`.  I believe I could use `via1_write()` instead.  Thoughts?

mist64: I cleaned up this part in the VIA in e805d14. Would it work now?

jjgignac: Much better.  Would you like me to rebase on the current master and crush into one commit?

PR #168, opened 2019-10-12
Use enum for GIF recorder.

rsbohn:


mist64:
Can you use the *_t type when declaring the variables?

PR #175, opened 2019-10-15
Fix minor joystick bug

jjbliss:
In cases where more than 24-bits of data are read from the joystick ports, the emulator was not acting as real hardware would.  I can't think of a real use case where this would be a problem, but this will bring it more into line with actual hardware.

Correct behavior:
After the data bits are written any additional bits from controllers are 0.
If no controller present, always write 1.

Problem behavior:
The emulator was sending any additional bits after 24 as 1, even when controllers are plugged in.
The correct behavior is to send in 0 when controllers are plugged in and 1 when they are not.

Fix:
This new code checks whenever joystick data is written and always writes a 1 if no joystick is set up, so that means we only need to actually generate the 16 bits for joystick state and just write 0 for any additional bits when joysticks are present.


greg-king5:
Now, `get_joystick_state()` should return `uint16_t`.

jjbliss:
> Now, `get_joystick_state()` should return `uint16_t`.

Thanks for pointing that out, it should be fixed now.

PR #182, opened 2019-10-16
Add Windows cross-compilation to Travis CI

tobier:
This should help with PRs by verifying that we can still cross-compile for Windows with mingw32-w64 as done in the Makefile already.


tobier:
Had an old branch and forgot to rebase, hence the multiple amends.

Issue #179, opened 2019-10-15
Missing control over TILEW and TILEH in text modes


Serentty:
The [official specifications list](https://www.facebook.com/groups/CommanderX16/permalink/402529400498160/) for the Commander X16 mentions that “the 1 bpp tile modes in combination with 8x16 tile size will also allow for a standard VGA font to be used.” By “1 bpp tile modes” I'm assuming that it's referring to modes zero and one, the text modes. However, currently the emulator ignores TILEW and TILEH in these two modes, as can be seen in the code below.

```c
if (props->tile_mode) {
    props->tilew = 1 << (((reg_layer[layer][1] >> 4) & 1) + 3);
    props->tileh = 1 << (((reg_layer[layer][1] >> 5) & 1) + 3);
} else {
    props->tilew = 8;
    props->tileh = 8;
}
```

I discovered this when I went to use this feature in order to display Japanese text at a reasonable size and resolution, yet nothing happened.

mist64:
I don't see this claim in the specification (Facebook is ^*&$^#%&!), and it's not in the VERA spec:

https://github.com/commanderx16/x16-docs/blob/master/VERA%20Programmer's%20Reference.md

@fvdhoef Can you comment?

fvdhoef:
It is correct that tile mode 0 and 1 also support all combinations of tile width and height 8x8, 8x16, 16x8, 16x16. If the emulator doesn't emulate this now, that is an error.

mist64:
A pull request would be gladly accepted. :)

Serentty:
Alright, if this is supposed to be emulated, I'll take a look and see if I can figure out how to implement it.

Kroc:
Can't access Facebook _at all_; can anybody perhaps mirror that doc please?

indigodarkwolf:
Hopefully the fix is straightforward, just remove the if() and use the tile size logic from tiles modes for text as well:

```C
props->tilew = 1 << (((reg_layer[layer][1] >> 4) & 1) + 3);
props->tileh = 1 << (((reg_layer[layer][1] >> 5) & 1) + 3);
```

But even if it's harder than that, I expect the worst case would be looking at wherever checks `props->tile_mode` and `props->text_mode`, which shouldn't be too bad.

Serentty:
[...]

PR #184, opened 2019-10-16
Remove checks on tile size for text modes (issue #179)

Serentty:
This removes the check preventing the text modes from setting a different tile size, allowing for 8×16, 16×8, and 16×16 characters.


Serentty:
Testing with the 8×16 VGA font seems to indicate that this works properly.

image

mist64:
Excellent, thank you!

PR #185, opened 2019-10-16
Video optimization (Linux CPU 100% usage, Issue #51)

neilogd:
- render_line:
  - Recalculating the XRGB color from palette per line only when changed, not when writing to the framebuffer.
  - Split up the loop over pixels in a line into discrete stages:
      1) Calculate color indices.
      2) Add border.
      3) Look up XRGB color & write to framebuffer.
      4) Apply NTSC overscan.
   - Calculate 8 color indices per iteration (was only maybe a 50ms improvement for quite a bit more complexity... probably be much nicer if flatted by hand to SIMD intrinsics)

- render_layer_line:
  - Calculate VRAM range to read in, and read in the range rather than several individual bytes when the range lies within VRAM, otherwise fallback to individual byte reads.

Timing video_step over 300 frames (158025 calls):
Baseline:
    2477.246284 ms total, 0.015676 ms avg,
    **8.257487613 ms/frame**

Plus optimized render_line:
    1848.739386 ms total, 0.011699 ms avg,
    **6.16246462 ms/frame**

Plus optimized render_layer_line:
    1673.707724 ms total, 0.010591 ms avg,
    **5.579025747 ms/frame**

Additional testing with @indigodarkwolf's x16-racer, using "-log S" specifically:
```
Before (Settles at 88-90%):
Load: 100%
Rendering is behind -58 frames.
Load: 100%
Rendering is behind -58 frames.
Load: 100%
Rendering is behind -58 frames.
Load: 100%
Rendering is behind -58 frames.
Load: 100%
Rendering is behind -59 frames.
Load: 100%
Rendering is behind -59 frames.
Load: 100%
Rendering is behind -59 frames.
Load: 100%
Rendering is behind -59 frames.
Load: 100%
Rendering is behind -59 frames.
```
After:
```
Load: 100%
Load: 88%
Load: 94%
Load: 94%
Load: 82%
Load: 100%
Load: 88%
Load: 88%
Load: 88%
Load: 100%
Load: 82%
Load: 100%
Load: 88%
Load: 94%
Load: 82%
```

Test CPU is an Intel(R) Core(TM) i5-5250U CPU @ 1.60GHz.

### DIFF
--- a/video.c
+++ b/video.c
@@ -11,6 +11,8 @@
 #include "gif.h"
 #include "vera_spi.h"
 
+#include <limits.h>
+
 #ifdef __EMSCRIPTEN__
 #include "emscripten.h"
 #endif
@@ -66,6 +68,10 @@
 #define RSHORTCUT_KEY SDL_SCANCODE_RCTRL
 #endif
 
+// When rendering a layer line, we can amortize some of the cost by calculating multiple pixels at a time.
+#define LAYER_PIXELS_PER_ITERATION 8
+
+
 static SDL_Window *window;
 static SDL_Renderer *renderer;
 static SDL_Texture *sdlTexture;
@@ -105,7 +111,9 @@ static const uint16_t default_palette[] = {
 };
 
 static uint8_t video_space_read(uint32_t address);
+static void video_space_read_range(uint8_t* dest, uint32_t address, uint32_t size);
 
+static void refresh_palette();
 
 void
 video_reset()
@@ -135,6 +143,8 @@ video_reset()
 		palette[i * 2 + 0] = default_palette[i] & 0xff;
 		palette[i * 2 + 1] = default_palette[i] >> 8;
 	}
+
+	refresh_palette();
 
 	scan_pos_x = 0;
 	scan_pos_y = 0;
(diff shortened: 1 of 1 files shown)


indigodarkwolf:
Looks like some more substantial wins. Very nice.

wg-nforbesrichardson:
I'd be tempted to push a bit more and squeeze some more juice, but at this stage in the project I'd rather wait until the hardware is unlikely to change much. Don't want to do optimization work that has a high chance of being obsolete in a revision or two if can be avoided :)

neilogd:
I'd be tempted to push a bit more and squeeze some more juice, but at this stage in the project I'd rather wait until the hardware is unlikely to change much. Don't want to do optimization work that has a high chance of being obsolete in a revision or two if can be avoided :)

mist64:
I'll get to this soon, sorry.

video.c
```diff
@@ -747,73 +873,130 @@ render_line(uint16_t y)
...
+		memset(spr_zindex, 0, sizeof(spr_zindex));
+
+		// Calculate color without border.
+		for (uint16_t x = 0; x < SCREEN_WIDTH; x+=LAYER_PIXELS_PER_ITERATION) {
+			uint8_t col_index[LAYER_PIXELS_PER_ITERATION];
+			memset(col_index, 0, sizeof(col_index));
+
+			int eff_x[LAYER_PIXELS_PER_ITERATION];
+			for (int i = 0; i < LAYER_PIXELS_PER_ITERATION; ++i) {
+				eff_x[i] = 1.0 / hscale * (x + i - hstart);
```
jkotlinski: `eff_x[i] = (reg_composer[1] * (x + i - hstart)) >> 7;` would be more efficient, since it avoids double division

jkotlinski: same trick goes for `eff_y`

mist64:
I'm planning to release r34 *without* this change later today (for stability), and have a look at this asap... :/

neilogd:
Ok cool. I figured you've had more important things to focus on for the mean time.

neilogd:
Sorry, had some issues with my github account so I think this PR went walkies for a week and a bit. Appears to be sorted out now.

mist64:
Thanks!

PR #188, opened 2019-10-18
skip IO area when loading PRG files

Frank-Buss:
This allows to load big PRG files, which then will be loaded
into the banked RAM area, if it reaches the end of the BASIC RAM
area. This allows easier program distribution without many small
files. Example app:
http://www.frank-buss.de/x16/mode7-full.prg
Both load functions are changed: -prg and when using LOAD from BASIC.
The native LOAD function with IEC will follow in the ROM repository,
when I found a way to test it.

### DIFF
--- a/loadsave.c
+++ b/loadsave.c
@@ -16,6 +16,8 @@
 #define MIN(a,b) (((a)<(b))?(a):(b))
 #define MAX(a,b) (((a)>(b))?(a):(b))
 
+extern uint16_t num_ram_banks;
+
 static int
 create_directory_listing(uint8_t *data)
 {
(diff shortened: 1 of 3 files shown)


kktos:
Frank, Just food for thought:
I suppose at some point we will have a way to tell if a bank is used or not. This will be part of the firmware.
So your code will have to deal with it. (ok, ok, when this will be settle, of course.)
And more: what if bank 0 is free, bank 1 used and  bank 2 free ?
The code will also have to get the available free banks....
But then, you'll tell me, how can I know the banks my program was loaded into......
yes, that's a good question.... :)

told you, just food for thought ;)

Frank-Buss:
Does the current ROM use the banked RAM? If not, I don't see a problem. Even if it uses it at some time, this doesn't matter much, if you can still load one PRG after reset for games etc. If the banked RAM is used, then the current implementation which can load directly to the banked RAM (without lower RAM) would be needed to change, too. But I think we shouldn't make it too difficult, it is not like a multitask, multiuser system :-)

kktos:
True, but I kept thinking we can load some extension, libraries (graphic), simon's basic, stuff like that. And it would be good if they are able to share the mem without a fight :)

Frank-Buss:
@mist64 So what about this PR? I'm writing a bigger demo at the moment with Barry, would be nice to know if I have to split it in 2 files, or if I could use just one file :-) Works with the native LOAD as well with [this PR](https://github.com/commanderx16/x16-rom/pull/73). Or it will work, when someday the SD-card implementation works, and the IEC implementation is finished, but at least while testing my implementation, I fixed another critical bug in the LOAD function.

PR #218, opened 2019-12-02
Commander X16 window icon

nigels-com:
Screenshot_20191202_202851

Based on screenshot, output to XPM from gimp, hand-tweaked for feeding to SDL surface.


mist64:
Merged as 014e944

Issue #202, opened 2019-11-04
Esc key doesn't exit BASIC program on R34


MCPEmaniac:
The Esc key doesn't seem to exit the following BASIC program on R34.

10 GOTO 10

mist64:
Workaround: CTRL+C

Esshahn:
PR submitted for fix: https://github.com/commanderx16/x16-emulator/pull/229

PR #229, opened 2019-12-30
fix for issue #202

Esshahn:
Moved define from video.c to keyboard.c, which seemed to be a cleanup
leftover.


nigels-com:
Looks good to me.

PR #233, opened 2020-01-05
IPad Version

mattuna15:
Add SDL dependency for iPad app
Add Xcode project and objective-c code
Add objective-c runtime calls into main emulator code


.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "xcode/SDL"]
+	path = xcode/SDL
+	url = https://github.com/mattuna15/SDL.git
```
mist64: This is a little scary.
* Can this be done without a submodule? Something like: For an iPad build, manually drop SDL.zip into this directory? Sure, hacky, but a submodule is quite a big hammer, and it's not needed for the other platforms.
* I see your SDL fork has a few changes. Why are they needed, and can we make this work with a vanilla SDL?

README.md
```diff
@@ -25,6 +25,16 @@ You can build a ROM image yourself using the [build instructions][x16rom-build]
 
 Install SDL2 using `brew install sdl2`.
 
+### iPad Build
+
+Using xcode open the project file in the emulator xcode directory.
```
mist64: Xcode

README.md
```diff
@@ -78,6 +88,8 @@ Run `x16emu -h` to see all command line options.
 Keyboard Layout
 ---------------
 
+* ON IPADS A LONG PRESS WILL GIVE YOU ACCESS TO THE FUNCTION KEYS
```
mist64: I'd rather have a special section that explains what's different on iPad.

debugger.c
```diff
@@ -12,7 +12,22 @@
 #include <stdlib.h>
 #include <stdio.h>
 #include <inttypes.h>
+
+#if __APPLE__
```
mist64: This is repeated a lot. Can we have something like this:

    #if __APPLE__ && (TARGET_IPHONE_SIMULATOR || TARGET_OS_IPHONE)
    #include "ios.h"
    #endif

and do the magic there?

Even if different .c files want different iOS headers to be included, we could just include all of them in "ios.h"; no harm in that.

Also, why special case Mac? What's "TargetConditionals.h"?

keyboard.c
```diff
@@ -224,42 +223,54 @@ ps2_scancode_from_SDL_Scancode(SDL_Scancode scancode)
 void
 handle_keyboard(bool down, SDL_Keycode sym, SDL_Scancode scancode)
 {
-	if (down) {
-		if (log_keyboard) {
-			printf("DOWN 0x%02X\n", scancode);
-			fflush(stdout);
-		}
+    if (down) {
+#if __APPLE__
```
mist64: again, combine into

    #if __APPLE__ && (TARGET_IPHONE_SIMULATOR || TARGET_OS_IPHONE)

keyboard.c
```diff
@@ -224,42 +223,54 @@ ps2_scancode_from_SDL_Scancode(SDL_Scancode scancode)
...
-				ps2_buffer_add(0, 0xe0);
-			}
-			ps2_buffer_add(0, ps2_scancode & 0xff);
-		}
-	} else {
-		if (log_keyboard) {
-			printf("UP   0x%02X\n", scancode);
-			fflush(stdout);
-		}
+            int ps2_scancode = ps2_scancode_from_SDL_Scancode(scancode);
```
mist64: The indentation changed here. In Xcode, you can change it to tabs by going to Preferences -> Text Editing -> Indentation. I think you can also do it by project somewhere.

loadsave.c
```diff
@@ -50,8 +54,23 @@ create_directory_listing(uint8_t *data)
 	*data++ = 'P';
 	*data++ = 'C';
 	*data++ = 0;
+    
+    char filename[256];
+    uint8_t len = MIN(RAM[FNLEN], sizeof(filename) - 1);
```
mist64: RAM[FNLEN] is a byte and can't be > 255, so the MAX() part is a no-op. I'd rather comment on the line above that 256 is the maximum possible KERNAL filename length + 1.

loadsave.c
```diff
@@ -50,8 +54,23 @@ create_directory_listing(uint8_t *data)
...
+    memcpy(filename, (char *)&RAM[RAM[FNADR] | RAM[FNADR + 1] << 8], len);
+    filename[len] = 0;
+    
+    //use correct file location for ios
+    #if __APPLE__ && (TARGET_IPHONE_SIMULATOR || TARGET_OS_IPHONE)
```
mist64: Now you're overwriting filename in all cases. The copying above had no effect.

(Seems correct, since the filename should be "$[...]" in this case, and can't be used as a filename/path on the host.)

loadsave.c
```diff
@@ -50,8 +54,23 @@ create_directory_listing(uint8_t *data)
...
+    strcpy(filename,getenv("HOME"));
+    //concatenating the path string returned from HOME
+    strcat(filename,"/Documents/.");
+    #else
+    strcpy(filename,".");
```
mist64: So I'd rather declare filename as a `char filename[256]` in the iOS path, and strcpy/strcat into it, and declare it as `const char *filename = "."` in the default path. This way, we don't allocate 256 bytes on the stack just to fill two bytes with constant data at runtime.

loadsave.c
```diff
@@ -132,7 +151,20 @@ LOAD()
 		RAM[STATUS] = 0;
 		a = 0;
 	} else {
-		FILE *f = fopen(filename, "rb");
+        
+        //use correct file location for ios
+        //use correct file location for ios
```
mist64: dup; also: move inside #ifdef

loadsave.c
```diff
@@ -132,7 +151,20 @@ LOAD()
...
+        #if __APPLE__ && (TARGET_IPHONE_SIMULATOR || TARGET_OS_IPHONE)
+        char fullFileName[256];
+        //concatenating the path string returned from HOME
+        strcpy(fullFileName,getenv("HOME"));
+        strcat(fullFileName,"/Documents/");
+        strcat(fullFileName,filename);
+        FILE *f = fopen(fullFileName, "rb");
+        #else
+        FILE *f = fopen(filename, "rb");
```
mist64: I'd suggest to do as I suggested above: `char full_filename[256]` and strcpy/strcat in the iOS case, `char *full_filename = filename` in the default case; then `FILE *f = fopen(full_filename, "rb");` to minimize code duplication between the paths.

loadsave.c
```diff
@@ -132,7 +151,20 @@ LOAD()
...
 		a = 0;
 	} else {
-		FILE *f = fopen(filename, "rb");
+        
+        //use correct file location for ios
+        //use correct file location for ios
+        #if __APPLE__ && (TARGET_IPHONE_SIMULATOR || TARGET_OS_IPHONE)
+        char fullFileName[256];
+        //concatenating the path string returned from HOME
+        strcpy(fullFileName,getenv("HOME"));
```
mist64: coding style: space after comma, please, to be consistent with the surrounding code.

main.c
```diff
@@ -423,6 +432,50 @@ void closeAudio()
 }
 #endif
 
+#if __APPLE__ && (TARGET_IPHONE_SIMULATOR || TARGET_OS_IPHONE)
```
mist64: It seems this doesn't do anything. Are some paths TODOs? If so, comments should be added. Otherwise, I'd remove the cases.

Maybe we want to do something about backgrounding at some point, like saving all state, in case we get killed – or are we guaranteed to get "terminating" in this case? In any case, we don't have full state save/restore yet, so that would be a TODO.

main.c
```diff
@@ -834,6 +887,13 @@ main(int argc, char **argv)
 
 	joystick_init();
 
+// first time through for iOS
+#if __APPLE__ && (TARGET_IPHONE_SIMULATOR || TARGET_OS_IPHONE)
+    SDL_SetEventFilter(HandleAppEvents, NULL);
+    createIosMessageObserver();
+    SDL_StartTextInput();
```
mist64: What does this do? Is it a hint to show the keyboard? Can we move this outside the #ifdef, because it would be generally useful for other touch-enabled platforms?

xcode/X16/iOS/GameControlViewController.h
```diff
@@ -0,0 +1,16 @@
+//
+//  GameControlViewController.h
+//  CommanderX16
+//
+//  Created by Pearce, Matthew (Senior Developer) on 03/01/2020.
```
mist64: Instead, can you write this:

    ; (C)2020 Matthew Pearce, License: 2-clause BSD

(in case you're ok with this license, otherwise reference your preferred license.)

debugger.c
```diff
@@ -12,7 +12,22 @@
...
+#include <TargetConditionals.h>
+#if TARGET_IPHONE_SIMULATOR || TARGET_OS_IPHONE
+#include "SDL.h"
```
mist64: if you add some magic path to your includes in Xcode, you shouldn't have to special case `#include "SDL.h"` over `#include <SDL.h>`

xcode/X16/iOS/GameControlViewController.m
```diff
@@ -0,0 +1,35 @@
...
+#include "SDL_keyboard.h"
+
+@interface GameControlViewController ()
+-(IBAction)buttonPressed:(id)sender;
+-(IBAction)closeWindow:(id)sender;
+@end
+
+@implementation GameControlViewController
+
+- (void)viewDidLoad {
```
mist64: delete

xcode/X16/iOS/MessageObserver.m
```diff
@@ -0,0 +1,58 @@
...
+    
+    return self;
+}
+
+-(void)showSettings {
+    
+    if (!settings) {
+        settings = [[SettingsViewController alloc] initWithNibName:@"SettingsViewController" bundle:nil];
+    } else {
+        [[[self rootView] viewWithTag:666] removeFromSuperview];
```
mist64: I know this is the first ObjC code in this codebase, so who am I to comment about coding style, but I'd *consistently* use dot notation for properties. Below, you say `settings.view.frame`, so you should say `self.rootView` here. And `window.rootViewController` and `delegate.window` below. (But keep `[SDLUIKitDelegate sharedAppDelegate]`, since it's not a property.)

xcode/X16/iOS/MessageObserver.m
```diff
@@ -0,0 +1,58 @@
...
+    }
+    
+    CGRect rect = settings.view.frame;
+    
+    CGPoint origin = CGPointMake(0, 0);
+    rect.origin = origin;
```
mist64: rect.origin = CGPointZero;

?

xcode/X16/iOS/MessageObserver.m
```diff
@@ -0,0 +1,58 @@
...
+    settings.view.frame = rect;
+    
+    settings.view.tag = 666;
+    
+    [[self rootView] addSubview:settings.view];
```
mist64: Why not `presentViewController:animated:completion:`? And is the else case above supposed to dismiss it? In which case, I'd suggest `dismissViewControllerAnimated:completion:`.

xcode/X16/iOS/SettingsViewController.m
```diff
@@ -0,0 +1,77 @@
...
+@implementation SettingsViewController
+
+@synthesize game;
+
+- (void)viewDidLoad {
+    [super viewDidLoad];
+    // Do any additional setup after loading the view from its nib.
+}
+
+/*
```
mist64: remove if not used

xcode/X16/iOS/SettingsViewController.m
```diff
@@ -0,0 +1,77 @@
...
+-(IBAction)showGamePad:(id)sender;
+
+@property (nonatomic, retain) GameControlViewController *game;
+@end
+
+@implementation SettingsViewController
+
+@synthesize game;
+
+- (void)viewDidLoad {
```
mist64: remove

xcode/X16/iOS/SettingsViewController.m
```diff
@@ -0,0 +1,77 @@
...
+// In a storyboard-based application, you will often want to do a little preparation before navigation
+- (void)prepareForSegue:(UIStoryboardSegue *)segue sender:(id)sender {
+    // Get the new view controller using [segue destinationViewController].
+    // Pass the selected object to the new view controller.
+}
+*/
+
+-(IBAction)toggleKeyboard:(id)sender {
+
+    SDL_StartTextInput();
```
mist64: You also do this in main(). Remove there? Remove here?

xcode/X16/iOS/SettingsViewController.m
```diff
@@ -0,0 +1,77 @@
...
+
+-(IBAction)buttonPressed:(id)sender
+{
+    UIButton *button = (UIButton *)sender;
+    SDL_SendKeyboardKey(1, button.tag);
+}
+
+-(IBAction)closeWindow:(id)sender
+{
+    [self.view removeFromSuperview];
```
mist64: `dismissViewControllerAnimated:completion:`?

xcode/X16/iOS/ios_functions.c
```diff
@@ -0,0 +1,51 @@
+//
```
mist64: yay, msgSend magic! :)

But I don't see why this can't be ios_functions.m and use ObjC? The interface to the main code can still be the same, using a C interface.

xcode/X16/iOS/GameControlViewController.xib
```diff
@@ -0,0 +1,99 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
mist64: XIBs are impossible to review. :( I'm trusting you this is complex enough that you don't want to build any of this in code... :)

.gitmodules
mattuna15: Sure I can do that.

debugger.c
mattuna15: The mac uses the standard code, TargetConditionals.h allows you to target specific apple platforms.

xcode/X16/iOS/SettingsViewController.m
mattuna15: You have to have a way of restarting input. If you completely dismiss the keyboard then SDL stops accepting input.

README.md
mattuna15: 👍

.gitmodules
mattuna15: I need to look at subclassing the SDL code in the main app.

xcode/X16/iOS/ios_functions.c
mattuna15: At the moment the observer of notifications sent between .c and SDL needs to be created through msgsend. I removed the notification msgsend though.

.gitmodules
mattuna15: sub module removed

mattuna15: Code changed now to use vanilla SDL included as a zip :)

keyboard.c
mattuna15: 👍

xcode/X16/iOS/GameControlViewController.h
mattuna15: 👍

main.c
mattuna15: 👍

main.c
mattuna15: :+1:

main.c
```diff
@@ -471,7 +488,7 @@ main(int argc, char **argv)
 			int kb = atoi(argv[0]);
 			bool found = false;
 			for (int cmp = 8; cmp <= 2048; cmp *= 2) {
-				if (kb == cmp)  {
+				if (kb == cmp)	{
```
mist64: Thanks. I had the butterfly keyboard for a while. :(

xcode/X16/iOS/GameControlViewController.xib
mattuna15: Hoping that these are going to be temporary.

xcode/X16/iOS/GameControlViewController.xib
```diff
@@ -1,99 +1,99 @@
...
-                    </connections>
-                </button>
-            </subviews>
-            <color key="backgroundColor" red="1" green="1" blue="1" alpha="1" colorSpace="custom" customColorSpace="sRGB"/>
-            <freeformSimulatedSizeMetrics key="simulatedDestinationMetrics"/>
-            <viewLayoutGuide key="safeArea" id="Q5M-cg-NOt"/>
-            <point key="canvasLocation" x="98" y="-4"/>
-        </view>
-    </objects>
+	<device id="ipad12_9" orientation="landscape" layout="fullscreen" appearance="light"/>
```
mist64: .xib with tabs? I'm sure Xcode will change it into spaces again the next time it saves it.

mattuna15: If you're happy with me adding them to the editor config then I'll do so.

xcode/X16/iOS/SettingsViewController.m
mattuna15: Need to look at how SDL manages windows and views. This is the quickest way at the moment. I'll maybe add this as a todo.

xcode/X16/iOS/MessageObserver.m
mattuna15: Yep - will go through tidying up as I need to change things. Been doing swift for so long now!

xcode/X16/iOS/MessageObserver.m
mattuna15: removed now.

xcode/X16/iOS/MessageObserver.m
mattuna15: Windows in SDL seem to be problematic. I need to investigate further but it works for the moment :)

main.c
mattuna15: 👍

loadsave.c
mattuna15: 👍

loadsave.c
mattuna15: 👍

debugger.c
mattuna15: 👍

mist64:
@mattuna15 Tell me when it's a good time to look over the change again.

mattuna15:
@mist64 will do. I'm working with Jimmy Dansbo to get a socket interface up and running. So I will try and include that. Thanks for the note.

mist64:
P.S.: I'd rather merge the current state than wait for all features. :)

mattuna15:
Great no problems.

mattuna15:
@mist64 good to go. 👍

loadsave.c
```diff
@@ -51,7 +57,17 @@ create_directory_listing(uint8_t *data)
 	*data++ = 'C';
 	*data++ = 0;
 
-	if (!(dirp = opendir("."))) {
+	#if __APPLE__ && (TARGET_IPHONE_SIMULATOR || TARGET_OS_IPHONE)
```
madebr: You do this concatening a few times in this file.
Maybe extract it to some static helper function?
You can move the ifdef to it too.

video.c
```diff
@@ -933,6 +932,14 @@ video_update()
 			mouse_x = event.motion.x;
 			mouse_y = event.motion.y;
 		}
+#if __APPLE__ && (TARGET_IPHONE_SIMULATOR || TARGET_OS_IPHONE)
+		if (event.type == SDL_DROPFILE) {	   // In case if dropped file
```
madebr: Can this drop event also be enabled on other systems?
I mean, it would also be nice to have this on my PC.

xcode/Frameworks/JoyStickView/JoyStickView.xcodeproj/xcuserdata/mpe18.xcuserdatad/xcschemes/xcschememanagement.plist
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
madebr: @mist64
Having support for a build system such as cmake or meson would obviate the need for inclusion of these project files.

loadsave.c
mattuna15: 👍 thanks I'll take a look.

video.c
mattuna15: Yes - sure I'll remove the conditional.

PR #237, opened 2020-01-18
Switch file operations to use SDL

### LINKED ISSUE #221: X16Emu doesn't run for Windows users with nonstandard names
In a fit of stupidity, I once decided my Windows username should be Björn. On running the exe, I get "Cannot open C:\Users\Bj├Ârn\Desktop\x16emu_win-r34\rom.bin!" and then back to prompt.

Serentty:
This is intended to fix #221, but it has the effect of fixing a few other issues as well, including inconsistency over the encoding of BASIC files. Before merging, it will be necessary to test that all functionality is still working, which is something I'm already working on.

Issue #221 seems fixed by this, as now it is possible to load the ROM from a path containing arbitrary characters.

### DIFF
--- a/loadsave.c
+++ b/loadsave.c
@@ -8,6 +8,7 @@
 #include <sys/stat.h>
 #include <dirent.h>
 #include <unistd.h>
+#include <SDL.h>
 #include "glue.h"
 #include "memory.h"
 #include "video.h"
@@ -132,15 +133,17 @@ LOAD()
 		RAM[STATUS] = 0;
 		a = 0;
 	} else {
-		FILE *f = fopen(filename, "rb");
+		SDL_RWops *f = SDL_RWFromFile(filename, "rb");
 		if (!f) {
 			a = 4; // FNF
 			RAM[STATUS] = a;
 			status |= 1;
 			return;
 		}
-		uint8_t start_lo = fgetc(f);
-		uint8_t start_hi = fgetc(f);
+		uint8_t start_lo;
+		SDL_RWread(f, &start_lo, sizeof(start_lo), 1);
+		uint8_t start_hi;
+        SDL_RWread(f, &start_hi, sizeof(start_hi), 1);
 		uint16_t start;
 		if (!RAM[SA]) {
 			start = override_start;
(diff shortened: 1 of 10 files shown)


loadsave.c
```diff
@@ -132,15 +133,17 @@ LOAD()
...
 			RAM[STATUS] = a;
 			status |= 1;
 			return;
 		}
-		uint8_t start_lo = fgetc(f);
-		uint8_t start_hi = fgetc(f);
+		uint8_t start_lo;
+		SDL_RWread(f, &start_lo, sizeof(start_lo), 1);
+		uint8_t start_hi;
+        SDL_RWread(f, &start_hi, sizeof(start_hi), 1);
```
greg-king5: I'm curious to know why you don't use `SDL_ReadU8()` in place of `fgetc()`, and `SDL_WriteU8()` in place of `fputc()`.

vera_uart.c
```diff
@@ -14,15 +15,28 @@ static uint8_t byte_in;
...
+SDL_RWops *uart_out_file = NULL;
 
 static bool
 txbusy()
 {
 	return countdown_out > 0;
 }
 
+bool
+check_eof(SDL_RWops *f)
```
greg-king5: That function should be defined as `static`.

loadsave.c
greg-king5: In fact, at this place in the code, you can use
```c
uint16_t start = SDL_ReadLE16(f);
```

Serentty:
Oh, so there _is_ a function what does the same thing as ``fgetc()``? I was looking through the SDL documentation and didn't see that because it wasn't listed under the functions for ``SDL_RWops``. I'll fix that and the static thing.

greg-king5:
> Oh, so there _is_ a function what does the same thing as `fgetc()`? I was looking through the SDL documentation, and didn't see that because it wasn't listed under the functions for `SDL_RWops`.

https://wiki.libsdl.org/CategoryIO

Serentty:
I'll get around to making those small fixes soon. I did all of this on a huge binge one night and need to wait until I have that motivation again.

mist64:
Superseded by https://github.com/commanderx16/x16-emulator/pull/256

Serentty:
@mist64 It seems I kept putting this off until it got taken care of anyway. Sorry about that, and thanks for your hard work!

PR #241, opened 2020-01-29
Corrected sound frequencies

MJoergen:
Playing a 440 Hz sound (by writing the value $4A to the YM2151 register $28) did not produce a sound output of the desired frequency, but rather 246 Hz. This was verified using both audacity as well as an external guitar tuner.

According to the YM2151 documentation, frequencies generated are based on the chip clock frequency, which is expected to be 3.579545 MHz.

For some (for me unknown) reason the value needed in line 468 is apparently twice the chip clock frequency. With this single change, the sound output now has the desired frequency, verified in both audacity and an external guitar tuner.

### DIFF
--- a/main.c
+++ b/main.c
@@ -464,8 +464,8 @@ init_audio()
 		exit(-1);
 	}
 
-	// init YM2151 emulation. 4 MHz clock
-	YM_Create(4000000);
+	// init YM2151 emulation. twice the 3.579545 MHz chip clock
+	YM_Create(7159090);
 	YM_init(have.freq, 60);
 
 	// start playback


fvdhoef:
I did some testing on my own using the VGM player of Frank Buss (see x16-demo/cc65-audio). If I compare the output of his player to that on the vgmrips site, the pitch is too high. So you're correct that the frequency should probably be 3.579545MHz for this. But when I pass 2 times the frequency to YM_Create, the pitch is way too high.

Elektron72:
I tested the emulation of the YM 2151 using the [YM 2151 Synth UI](https://www.commanderx16.com/forum/index.php?/files/file/44-ym2151-synth-ui/), and after adding some log messages to the emulator to check what was being sent to the YM2151 emulation functions, I believe that the correct value for line 468 is 3579545. For me, this value gave an output frequency of about 440 Hz with register $28 set to $4A.

ZeroByteOrg:
I hope this gets pulled into the master branch because this discrepancy has been bugging the ever-living hell out of me. I know why it was set to 4MHz originally (Kevin's original test board used a 4MHz clock) but now that it's a well-known fact that the final system is going to use 3.5MHz, it's high time this got fixed - lol.

ZeroByteOrg:
There's no way the freq should be set to 2x 3.5MHz. I know the YM2151 datasheet specifies OCT=4, KC=A to deliver a 440Hz tone, but this is probably an error in the document (there are many). I play with VGM tunes all the time in emulated cores, and the X16-Emulator is only one tone sharp. Raising the clock speed to 2x reference clock would make all existing data play back an entire octave too high.

ZeroByteOrg:
Per Michael Steil's request on the commanderx16.com forums, I'm commenting that the correct clock rate for YM2151 in the Commander X16 is 3.579545 MHz. (See Elektron72's comment above as well)

indigodarkwolf:
Ultimately, though, the question is whether the YM2151 chip on the hardware is being run at 8MHz (same clock as the CPU) or closer to 3.579545MHz (maybe some division of the VERA clock). @fvdhoef @mist64 Does anyone know the answer to this?

Elektron72:
It was confirmed to be 3.579545 MHz at [https://www.commanderx16.com/forum/index.php?/topic/1210-proto-2-support-in-github-rom-emulator-documentation/&do=findComment&comment=9157](https://www.commanderx16.com/forum/index.php?/topic/1210-proto-2-support-in-github-rom-emulator-documentation/&do=findComment&comment=9157).

indigodarkwolf:
I didn't take Michael's reply to confirm that we knew the clock rate of the chip on the physical board. Seeing as he's the only one with the hardware, though, I suppose he's the only one who would know if the current emulator's specified clock speed is wrong.

indigodarkwolf:
Well, submitted PR #339 to correct the frequency. Can't update the code in this PR because I don't have write permissions to the branch. If #339 is accepted, we should close/reject this PR.

MJoergen:
As others have noted, this PR is incorrect. Since there are now other PR's with the correct frequency, I'm closing this PR.

PR #256, opened 2020-03-22
Switch file operations to use SDL

mist64:
Replaces pull request #237


Serentty:
I've learned since this that Windows has finally added a feature that allows programs to choose their own codepage, and it only took them 35 years. It's as easy as setting a field in the EXE's manifest. However, only Windows 10 1903 and later versions know about this field, so the solution that we went with is more compatible.

PR #258, opened 2020-03-24
Added sprite collision interrupt

indigodarkwolf:
This adds sprite collision detection and interrupt generation to video.c. The kernal does not clear the sprite collision interrupt flag, so examples in BASIC will appear to hang as soon as sprite pixels overlap. See also the sample programs [linked here](https://github.com/indigodarkwolf/x16-samples/tree/master/sprites-collision).


indigodarkwolf:
Oh, I notice that I forgot to remove the non-GNU version of the ternary operators. If that's a problem, we can return those to the GNU-only version.

PR #272, opened 2020-05-14
Speed Up Rendering

mikey-b:
**Extra compiler flags for better output.**
The options -march=native -mcpu=native -mtune=native -DNDEBUG -flto -fexpensive-optimizations yeilds better results on more limited hardward (e.g. raspberry pi). Less noticable differences on more powerful hardware, but worth having. Compile time does increase but the code is small enough not to be an issue.

Added max(), min() macros
**Heavy Video.c refactoring.**
Some big changes here that need a proper review.

-  Remove LAYER_PIXELS_PER_ITERATION

My measurements have seen no benefits to having this loop unrolling in the code.

- Line 456 - Break instead of continue. We can just break out of this loop as line_x will always be larger than SCREEN_WIDTH once tested as such?

- Seperate the three line_render "modes". Text, Bitmap, and.. something else? These are much clearer seperated, and removes unnessesary calculations.

- NTSC and Border. Needs better test cases to make sure it still works corectly, these changes prob not needed.

### DIFF
--- a/video.c
+++ b/video.c
@@ -607,38 +607,6 @@ render_layer_line_tile(uint8_t layer, uint16_t y)
 	uint32_t y_add = 0;
 
 	int tile_change_index = 0;
-
-	inline void calc_tile_change_index_iteration(int eff_x) {
-		// extract all information from the map
-		uint32_t map_addr = calc_layer_map_addr_base2(props, eff_x, eff_y) - map_addr_begin;
-		tile_change_index = 16;
-
-		uint8_t byte0 = tile_bytes[map_addr];
-		uint8_t byte1 = tile_bytes[map_addr + 1];
-
-		// Tile Flipping
-		vflip = (byte1 >> 3) & 1;
-		hflip = (byte1 >> 2) & 1;
-
-		yy = eff_y & props->tileh_max;
-		if (vflip) {
-			yy = yy ^ (props->tileh_max);
-		}
-		if (hflip) {
-			hflip_value = (props->tilew_max);
-		} else {
-			hflip_value = 0;
-		}
-		// additional bytes to reach the correct line of the tile
-		y_add = (yy << (props->tilew_log2 + props->color_depth)) >> 3;
-
-		tile_index = byte0 | ((byte1 & 3) << 8);
-		// offset within tilemap of the current tile
-		tile_start = tile_index << props->tile_size_log2;
-		tile_offset = props->tile_base + tile_start + y_add;
-
-		palette_offset = byte1 & 0b11110000;
-	}
 	// -------------------
 
 	// Render tile line.
@@ -648,7 +616,35 @@ render_layer_line_tile(uint8_t layer, uint16_t y)
 		int xx = eff_x & props->tilew_max;
 
 		if (tile_change_index == 0) {
-			calc_tile_change_index_iteration(eff_x);
+			// extract all information from the map
+			uint32_t map_addr = calc_layer_map_addr_base2(props, eff_x, eff_y) - map_addr_begin;
+			tile_change_index = 16;
+
+			uint8_t byte0 = tile_bytes[map_addr];
+			uint8_t byte1 = tile_bytes[map_addr + 1];
+
+			// Tile Flipping
+			vflip = (byte1 >> 3) & 1;
+			hflip = (byte1 >> 2) & 1;
+
+			yy = eff_y & props->tileh_max;
+			if (vflip) {
+				yy = yy ^ (props->tileh_max);
+			}
+			if (hflip) {
+				hflip_value = (props->tilew_max);
+			} else {
+				hflip_value = 0;
+			}
+			// additional bytes to reach the correct line of the tile
+			y_add = (yy << (props->tilew_log2 + props->color_depth)) >> 3;
+
+			tile_index = byte0 | ((byte1 & 3) << 8);
+			// offset within tilemap of the current tile
+			tile_start = tile_index << props->tile_size_log2;
+			tile_offset = props->tile_base + tile_start + y_add;
+
+			palette_offset = byte1 & 0b11110000;
 		}
 		tile_change_index -= 1;
 
--- a/video.c
+++ b/video.c
@@ -618,7 +618,7 @@ render_layer_line_tile(uint8_t layer, uint16_t y)
 		if (tile_change_index == 0) {
 			// extract all information from the map
 			uint32_t map_addr = calc_layer_map_addr_base2(props, eff_x, eff_y) - map_addr_begin;
-			tile_change_index = 16;
+			tile_change_index = props->tilew;
 
 			uint8_t byte0 = tile_bytes[map_addr];
 			uint8_t byte1 = tile_bytes[map_addr + 1];
--- a/video.c
+++ b/video.c
@@ -17,21 +17,6 @@
 
 #include <limits.h>
 
-#define max(a,b)             \
-({                           \
-    __typeof__ (a) _a = (a); \
-    __typeof__ (b) _b = (b); \
-    _a > _b ? _a : _b;       \
-})
-
-#define min(a,b)             \
-({                           \
-    __typeof__ (a) _a = (a); \
-    __typeof__ (b) _b = (b); \
-    _a < _b ? _a : _b;       \
-})
-
-
 #ifdef __EMSCRIPTEN__
 #include "emscripten.h"
 #endif
@@ -78,7 +63,7 @@
 #endif
 
 // When rendering a layer line, we can amortize some of the cost by calculating multiple pixels at a time.
-// #define LAYER_PIXELS_PER_ITERATION 8
+#define LAYER_PIXELS_PER_ITERATION 8
 
 
 static SDL_Window *window;
@@ -108,6 +93,8 @@ static uint8_t reg_composer[8];
 static uint8_t layer_line[2][SCREEN_WIDTH];
 static uint8_t sprite_line_col[SCREEN_WIDTH];
 static uint8_t sprite_line_z[SCREEN_WIDTH];
+static bool layer_line_enable[2];
+static bool sprite_line_enable;
 
 float scan_pos_x;
 uint16_t scan_pos_y;
@@ -232,28 +219,29 @@ struct video_layer_properties
 	uint16_t hscroll;
 	uint16_t vscroll;
 
-	uint16_t mapw;
-	uint16_t maph;
+	uint8_t  mapw_log2;
+	uint8_t  maph_log2;
 	uint16_t tilew;
 	uint16_t tileh;
+	uint8_t  tilew_log2;
+	uint8_t  tileh_log2;
 
 	uint16_t mapw_max;
 	uint16_t maph_max;
 	uint16_t tilew_max;
 	uint16_t tileh_max;
-
-	uint16_t layerw;
-	uint16_t layerh;
-
 	uint16_t layerw_max;
 	uint16_t layerh_max;
 
-	uint8_t bits_per_pixel;
-
-	uint32_t tile_size;
+	uint8_t tile_size_log2;
 
 	int min_eff_x;
 	int max_eff_x;
+
+	uint8_t bits_per_pixel;
+	uint8_t first_color_pos;
+	uint8_t color_mask;
+	uint8_t color_fields_max;
 };
 
 struct video_layer_properties layer_properties[2];
@@ -270,11 +258,19 @@ calc_layer_eff_y(struct video_layer_properties *props, int y)
 	return (y + props->vscroll) & (props->layerh_max);
 }
 
-static inline uint32_t
-calc_layer_map_addr(struct video_layer_properties *props, int eff_x, int eff_y)
+static uint32_t
+calc_layer_map_addr_base2(struct video_layer_properties *props, int eff_x, int eff_y)
 {
-	return props->map_base + (eff_y / props->tileh * props->mapw + eff_x / props->tilew) * 2;
+	// Slightly faster on some platforms because we know that tilew and tileh are powers of 2.
+	return props->map_base + ((((eff_y >> props->tileh_log2) << props->mapw_log2) + (eff_x >> props->tilew_log2)) << 1);
 }
+
+//TODO: Unused in all current cases. Delete? Or leave commented as a reminder?
+//static uint32_t
+//calc_layer_map_addr(struct video_layer_properties *props, int eff_x, int eff_y)
+//{
+//	return props->map_base + ((eff_y / props->tileh) * props->mapw + (eff_x / props->tilew)) * 2;
+//}
 
 void refresh_layer_properties(uint8_t layer)
 {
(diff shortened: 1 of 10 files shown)


indigodarkwolf:
I like most of these changes, personally. Though it looks like several of the platforms don't like the -fexpensive-optimizations flag.

video.c
```diff
@@ -1,1185 +1,1226 @@
...
+	// If video output is enabled, calculate color indices for line.
+	if (out_mode != 0) {
+		uint8_t spr_col_index = 0;
+		uint8_t l1_col_index = 0;
+		uint8_t l2_col_index = 0;
+		uint8_t spr_zindex = 0;
+
+		// Calculate color without border.
+		for (uint16_t x = 0; x < SCREEN_WIDTH; x += 1) {
+			const int eff_x = (reg_composer[1] * (x - hstart)) / 128; // What is this 128, should this be SCREEN_WIDTH?
```
rossato, line 793: The 128 here is the horizontal scaling factor.  A reg_composer[1] of 128 equals 1 input pixel to 1 output pixel scaling, 64 equals 1:2, and so on.

greg-king5:
None of the new optimization flags can be used; they aren't portable.  Different platforms use different CPUs and different compilers.

sebastianvog:
I did test/compile this for the WebAssembly target (without the Makefile changes):
Running the game ChaseVault as an example: On an older 2012 MacMini, I am seeing +10-20% improvement(for the % displayed in the title bar)
.|Chrome | Firefox | Safari
---------|------------ | ------------- | ----
Current|77% | 75% | 88% 
Optimized|89% | 95% | 100% 

It made a big impact in playability.

indigodarkwolf:
> 
> 
> If we "continue", line_x will still be larger than screen_width.

This is incorrect. In the event that a sprite's x coordinate (which is a signed number) is less than 0, the `uint16_t line_x` can underflow into large values, and becomes much greater than `SCREEN_WIDTH`.

Since sprite coordinates are based on the upper-left corner of the sprite, a sprite with an x coordinate less than 0 might be wholly off-screen, or only partially off-screen.

If we wanted to break, we would need to change line_x to a signed type, and an additional test `if(line_x < 0) continue;` I'm not sure if the extra test is worthwhile in the average case, both forms of the test have different optimal and degenerate cases. I suppose I'd rather optimize for fewer tests, on the theory that most people will store sprites vertically offscreen instead of horizontally offscreen, or will disable unused sprites entirely with a z-depth of 0.

indigodarkwolf:
`memset(col_line + hstop, border_color, min(hstart, SCREEN_WIDTH));`

That's `uint8_t border_color`, casting from 8-bit to 32-bit is not going to produce the results you expect...

indigodarkwolf:
So, I like some of these changes more, but sans profiling data (particularly on problematic platforms) I like others less. 

To that end, could I get someone to compare performance with https://github.com/indigodarkwolf/x16-emulator/tree/r37-optimizations on webasm targets and/or Raspberry Pis, specifically? Any non-Windows platforms would be informative, but webasm and RPis seem like the current "real" targets to work on improving.

That fork specifically splits the line drawing into three functions, like @mikey-b's changes, which does seem to optimize really well, but also focuses on removing a bunch of multiplies, divides, and modulos and replacing them with bit-shifts, based on much of the data always being powers-of-two. I also pulled a couple of piece of math out of loops that didn't need to be re-calculated. I've tested with my own assembly samples ([Matriculate](https://github.com/indigodarkwolf/x16-matriculate-text) and [X16-Racer](https://github.com/indigodarkwolf/x16-racer)), as well as @SlithyMatt's [Chase Vault](https://github.com/SlithyMatt/x16-chasevault), so I'm feeling fairly confident that it's correct.

The remaining hotspot that shows up in my profiling is this one line in `render_layer_line_tile`:
```C
		// convert tile byte to indexed color
		uint8_t col_index = (s >> (props->first_color_pos - ((xx & props->color_fields_max) << props->color_depth))) & props->color_mask;
```

It replaces the following block:
```C
			// convert tile byte to indexed color
			if (props->bits_per_pixel == 1) {
				col_index = (s >> (7 - (xx & 7))) & 1;

				if (props->text_mode) {
					col_index = col_index ? fg_color : bg_color;
				}
			} else if (props->bits_per_pixel == 2) {
				col_index = (s >> (6 - ((xx & 3) << 1))) & 3;
			} else if (props->bits_per_pixel == 4) {
				col_index = (s >> (4 - ((xx & 1) << 2))) & 0xf;
			} else if (props->bits_per_pixel == 8) {
				col_index = s;
			}
```

I'm not sure if there's anything more to be done about that, I suspect that additional gains will come from platform-specific caching strategies and threading.

sebastianvog:
> So, I like some of these changes more, but sans profiling data (particularly on problematic platforms) I like others less.
> 
> To that end, could I get someone to compare performance with https://github.com/indigodarkwolf/x16-emulator/tree/r37-optimizations on webasm targets 

I compiled the webassembly target with your r37-optimizations.

Here the results running ChaseVault.

.|Chrome | Firefox | Safari
---------|------------ | ------------- | ----
Current|77% | 75% | 88% 
mikey-b |89% | 95% | 100% 
indigo-d-w|100%|100%|100%

I see 100% or more (there is no percentage displayed once its over 99%) on all 3 browser. 

Very nice improvements everyone !

mikey-b:
Hi all !

Thank you for the comments, sorry I closed the request before seeing the comments. 

indigodarkwolf - Thank you for the time you've put in. You seem to have far better knowledge then me on the hardware and emulation requirements. I will continue to work on it, but will rebase to your fork and pass the changes up to yourself if thats OK.

> 
> 
> None of the new optimization flags can be used; they aren't portable. Different platforms use different CPUs and different compilers.

This all started with someone requesting a faster emulator on their raspberry pi, I wasn't aware that other targets were also a bit slow. 

The binaries might not be portable, but I think we should offer a "make native" option for those compiling there own? I also think we should be doing a windows x64 build as standard. so -m64 as a bare minimum?

indigodarkwolf:
Well, I'm definitely learning as well. As far as I was aware, previously the x86/x64 builds were reasonably optimized (given how much pipelining the chips do, and the herculean effort the chips themselves already go through to overcome costs from branching), and I didn't have any thoughts on where to find real gains, asides from threading.

I had no idea that ARM chips might struggle with integer division, and that tidbit from the Facebook thread got me to look at a whole new class of problems. Conveniently, that seems to have improved the performance of x86/x64 as well. 

At which point, "Cool." There's more to be plumbed, I'm sure.

indigodarkwolf:
Something I would add:

In your original diffs, there seemed to be a lot of redundant lines flagged as having changed. I suspect that had to do with newline characters. Please be more careful about this moving forward! It's just a real pain to identify the actual changes in a file when newlines have flagged _every single_ line.

mikey-b:
> 
> 
> > So, I like some of these changes more, but sans profiling data (particularly on problematic platforms) I like others less.
> > To that end, could I get someone to compare performance with https://github.com/indigodarkwolf/x16-emulator/tree/r37-optimizations on webasm targets
> 
> I compiled the webassembly target with your r37-optimizations.
> 
> Here the results running ChaseVault.
> . 	Chrome 	Firefox 	Safari
> Current 	77% 	75% 	88%
> mikey-b 	89% 	95% 	100%
> indigo-d-w 	100% 	100% 	100%
> 
> I see 100% or more (there is no percentage displayed once its over 99%) on all 3 browser.
> 
> Very nice improvements everyone !



> 
> 
> Something I would add:
> 
> In your original diffs, there seemed to be a lot of redundant lines flagged as having changed. I suspect that had to do with newline characters. Please be more careful about this moving forward! It's just a real pain to identify the actual changes in a file when newlines have flagged _every single_ line.

Thank you for the feedback - I have pulled back on the over-the-top stuff, and only changed render_line_bitmap() and render_line_tile(). I'm not sure if the diffs are clearer as its quite a bit moving? But taken your comment into consideration. 

My aim is to get the bottleneck to be data reads/writes. This is where we are after the attached suggested changes for both functions;

render_layer_line_tile()
1.5% | 662 | uint8_t col_index = (s >> (props->first_color_pos - ((xx   & props->color_fields_max) << props->color_depth))) &   props->color_mask;
1.0% | 666 | layer_line[layer][x] =   col_index;
0.6% | 658 | uint16_t x_add       = (xx << props->color_depth)   >> 3;

render_layer_line_bitmap()
0.9% | 694 | uint8_t col_index = (s >> (props->first_color_pos - ((xx   & props->color_fields_max) << props->color_depth))) &   props->color_mask;
0.7% | 691 | uint8_t s =   video_space_read(props->tile_base + tile_offset);
0.6% | 698 | layer_line[layer][x] =   col_index;

So yes, that col_index formula. Is there anything we can do to get some of it pre-computed and out of the loop?

**The changes suggested**
Benchmark before (out of 59.646s)
render_line 5.698s 
render_layer_line_bitmap 4.596s
render_layer_line_tile 3.909s

Benchmark with changes (out of 59.705s)
render_line 6.712s
render_layer_line_tile 2.625s
render_layer_line_bitmap 1.834s

OK so the current patch moves a lot of the props-> reads out of the loops. These *might* be moved out by the compiler - but due to aliasing some are clearly (as benchmarks show) not being moved. 

Palette_offset branch moved to outside of loop - Because we're capping col_index with & props->color_mask - We just check this is smaller than 16 (moving the 0 < col_index <16 test). Please check my thinking.

render_layer_line_tile needs a good reviewing. Looking at all the variables, a good chunk of them only seem to change every 16 iterations. So i've moved them out of the loop (into the lamdba calc_tile_change_index_iteration, updating them every 16'th iteration.

Is there a reason its 16? I suspect this may be a property of tiles? And may need to be read from somewhere (a layer property?), or is this a constant?

I think _bitmap is good to go, but _tile needs the above confirming. And anything that can be done with col_index = (s >>.. would be a bonus

mikey-b:
Removed lamdba syntax from render_line_tile as MacOS and WebAsm didn't accept it.

indigodarkwolf:
> Is there a reason its 16? I suspect this may be a property of tiles? And may need to be read from somewhere (a layer property?), or is this a constant?

You were in a 16-pixel width tile mode. This is what an 8-pixel width tile mode looks like:
https://imgur.com/a/CT1Unyh

(The glitch in the road is "intentional", it is not your fault. The background forest and mountains, on the other hand...)

mikey-b:
> 
> 
> > Is there a reason its 16? I suspect this may be a property of tiles? And may need to be read from somewhere (a layer property?), or is this a constant?
> 
> You were in a 16-pixel width tile mode. This is what an 8-pixel width tile mode looks like:
> https://imgur.com/a/CT1Unyh
> 
> (The glitch in the road is "intentional", it is not your fault. The background forest and mountains, on the other hand...)

Ok, do you know off hand if there is anywhere convenient this is stored for me to read in?

indigodarkwolf:
props->tilew

mikey-b:
> 
> 
> props->tilew

Stellar thank you! Changed to account for this on line 621

mikey-b:
> uint8_t col_index = (s >> (props->first_color_pos - ((xx & props->color_fields_max) << props->color_depth))) & props->color_mask;

I have managed to figure out that the shift to s is linearally progressing. And can be replaced with;

_Outside the loop:_
	uint8_t col_shift = props->first_color_pos;

	//if (props->color_depth == 3) { index_step = 0;
	uint8_t col_index_step = 0;
	if (props->color_depth == 2) {
		col_index_step = 4;
	} else if (props->color_depth == 1) {
		col_index_step = 2;
	} else if (props->color_depth == 0) {
		col_index_step = 1;
	}

_inside the loop:_
uint8_t col_index = (s >> col_shift) & props->color_mask;
col_shift = (col_shift - col_index_step) & 0b111;

This does improve _tile and _bitmap and we are now data read/write bottlenecked. The gain does not translate into extra frames, But i intend looking into speeding up composition function `render_line()`. So i have left the col_index adjustment out of this patch.

indigodarkwolf:
Then may I suggest replacing the if-chain with `const uint8_t col_index_step = props->bits_per_pixel`? Because that if-chain is literally just computing `1 << props->color_depth`, and that's literally already assigned to props->bits_per_pixel.

mikey-b:
> 
> 
> Then may I suggest replacing the if-chain with `const uint8_t col_index_step = props->bits_per_pixel`? Because that if-chain is literally just computing `1 << props->color_depth`, and that's literally already assigned to props->bits_per_pixel.

Excellent suggestion.

I have been looking elsewhere in the code base. 

```
new_frame |= video_step(MHZ);
...
video_step(float mhz) {
...
float advance = ((out_mode & 2) ? NTSC_PIXEL_FREQ :  VGA_PIXEL_FREQ) / mhz;
```

Because we're building them in seperate compilcation units - even though mhz is a constant, It can't be removed. So we've got a stack allocation, and a divide. 

Removing the parameter and using MHZ directly gives me 3% more frames. Best gain for a while, so i've added this to this current patch too.

mikey-b:
Infact a unity build gives me an extra 3% on top. We should maybe consider this for the build script?

`gcc -std=c99 -O3 -I/usr/include/SDL2 -D_REENTRANT -Iextern/include -Iextern/src -DNDEBUG cpu/fake6502.c *.c extern/src/ym2151.c -o x16emu_x -lSDL2 -lm -flto`

I have the Raspberry Pi build at 34% - from r37 which is 19%

mist64:
Sorry for the delay. I'm finally going through open pull requests for the emulator and the ROM at this point. What is the consensus on this, @mikey-b and @indigodarkwolf? Is this ready for merging?

mikey-b:
Hi Michael,

No it's not, a bug was found recently - Stephens version seems to be stable
and offers a decent speed up.

I am happy to help more if I can be of assistance - I was half expecting
some changes needed due to the Vera changes? The simulator really needs
unit tests IMHO, something im happy to work on too.

If you're just after a quick speed up win, I'd recommend applying Stephens
branch.

Kind regards,
Mike Brown

indigodarkwolf:
I'll organize my stuff into a proper PR, then, rather than just a ramshackle branch for active work.

mist64:
@mikey-b: I just merged @indigodarkwolf's speedup pull request. Are there any more speedup ideas in this one that we should continue considering?

Unit tests: That would be very very valuable!

mikey-b:
> 
> 
> @mikey-b: I just merged @indigodarkwolf's speedup pull request. Are there any more speedup ideas in this one that we should continue considering?
> 
> Unit tests: That would be very very valuable!

It depends on how much change you're willing to accept. Needs a bit of a refactoring IMHO, but maybe beyond the scope or willingness of the project.

There are two suggestions though that are easy changes

- Remove the MHZ argument from the rendering loop. Its known at compile time, and only used with other compile time values. The compiler can't remove it as it's currently compiled. Was this put in to allow for adjustable rendering?

- A unity build, to let the compiler do more with the code.

greg-king5:
> There are two suggestions, though, that are easy changes
> - Remove the MHZ argument from the rendering loop. Its known at compile time, and only used with other compile time values. The compiler can't remove it as it's currently compiled. Was this put in to allow for adjustable rendering?

I, for one, want a "warp speed" mode in the emulator!  If I run a program that will need a long time to finish, then I want to be able to speed up the emulation -- waiting all day isn't good for debugging.

mikey-b:
Hi Greg,

That's fair enough - I'm guessing the CPU clock is tied to the video? I
can't see any references to MHZ in the CPU to keep them in lockstep if
changed?

MHZ is currently a compile time constant (#define), It's just we're
converting it into a stack allocated variable, then handing it over to the
video code, which isn't necessary unless it was intended for this to be
dynamically changeable.

indigodarkwolf:
That being said, "warp speed" from the emulator simply means it would stop capping its processing based on framerate. If you want to see what "warp speed" could look like on the emulator, simply launch with the -debug command line option, hit F12, and wait a few minutes. Then hit F5 and watch the emulator run as fast as it possibly can until it catches up with the where it thinks it ought to be.

Note that if the emulator runs at anything less than 100% for you, you're *already* seeing it run as fast as possible, and no turbo button or control will be able to address that. "Warp speed" is definitely for the enthusiast PC crowd, unless or until we can find new ways to get substantial speed-ups from the video emulation, or else implement multithreading to fudge the performance across however many cores your system has available.

mist64:
"how much change you're willing to accept" – it's more of a question of how much speed we would gain. Single digit percentages are not worth making the code less readable/maintainable, for example.

MHZ: We don't need the MHZ at runtime, it can become a compile time constant in video_step() again. For warp mode (which I would find very useful!), all we would need to do is make the usleep() in emulator_loop() conditional.

mist64:
I just added warp mode in 9cc4d41. It can be improved a lot by skipping frames avoiding VERA rendering work.

mikey-b:
On Thu, 9 Jul 2020, 09:44 Michael Steil, <notifications@github.com> wrote:

> "how much change you're willing to accept" – it's more of a question of
> how much speed we would gain. Single digit percentages are not worth making
> the code less readable/maintainable, for example.
>
My guesstimate: with the current code base architecture, and with my
current experimental changes now bottlenecked on SPI logic - 10% may be
achieved, 100%+ is not likely.

From my understanding of other bit accurate simulation. This is normal
speeds.

But, we can confirm with a causal profiler. This'll give us a graph saying
"if you speed up this part of the code by this much, overall you'll go this
much faster". If it says speed up all these by 1000x to get 2x we know
we've maxed out.

> MHZ: We don't need the MHZ at runtime, it can become a compile time
> constant in video_step() again. For warp mode (which I would find very
> useful!), all we would need to do is make the usleep() in emulator_loop()
> conditional.
>
Stephens patch should get a lot of people close to, or over the line, to
emulating in real-time.

What would be your usecase for warp mode? It doesn't seem wise for
interactive use, for debug and automatic testing ?

If so, a debug mode frameskip might be what you want? There's a lot of
rendering code that doesn't alter the machine state if you are happy not
having input (I don't know how you would at high speeds?) and limited
output controls?

mikey-b:
> 
> 
> I just added warp mode in [82eabd8](https://github.com/commanderx16/x16-emulator/commit/82eabd841d2f4b3ddc6038072c48dc7d23a2da5e). It can be improved a lot by skipping frames avoiding VERA rendering work.

Just seen this, OK, looks like you have it in hand. 

I have been running benchmarks on the whole project that has shown some areas of possible improvement, but will put in a new thread. I think its safe to close this pull request.

PR #273, opened 2020-05-15
IRQ_LINE register can now be written

sasq64:


sasq64:
Ok never mind, there is already a pull request for this, that is better :)

PR #276, opened 2020-07-07
Replacing mults and divs with power-of-two adds.

indigodarkwolf:
This should be a properly reconciled set of changes with my previous PR, tested with ChaseVault, my X16 demos, and the sample sprite collide samples I wrote in https://github.com/indigodarkwolf/x16-samples/tree/master/sprites-collision.

### DIFF
--- a/video.c
+++ b/video.c
@@ -93,8 +93,10 @@ static uint8_t reg_composer[8];
 static uint8_t layer_line[2][SCREEN_WIDTH];
 static uint8_t sprite_line_col[SCREEN_WIDTH];
 static uint8_t sprite_line_z[SCREEN_WIDTH];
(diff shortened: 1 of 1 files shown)


indigodarkwolf:
I don't know why it insisted that I re-resolve things by hand, but whatever.

mist64:
Awesome. I'm measuring a speedup of about 20% on macOS.

sasq64:
Great indeed, a lot of changes and I very much doubt it was replacing muls with shifts that caused the speed up.

indigodarkwolf:
Well, not entirely, anyways, and certainly not on Intel platforms. This
also removed a lot of branching, which was probably the dominant
contributor.

PR #279, opened 2020-07-08
Adding VERA dump and memory fill commands to debugger commands

indigodarkwolf:
This adds the commands "v" and "f" to the debugger commands.

"v" follows the same syntax as "m", but changes the memory dump from system RAM/ROM to the VERA's VRAM. Using "m" returns the memory dump to showing system RAM/ROM.

Example syntax: 

Show VRAM, starting at address $10000: `v 10000`
Show VRAM, starting at address $00000: `v 0`

"f" fills memory starting at some address with a specified value. It expects up to 4 values: hex_addr hex_value dec_qty dec_incr:

- hex_addr is the start address to fill from, in hexadecimal
- hex_value is the 1-byte value to write to memory, in hexadecimal
- dec_qty is the number of times to write the hex_value to memory, and defaults to 1 if omitted
- dec_incr is how much to increment the target address by while filling, and defaults to 1 if omitted

Example syntax:

Fill $02 with $ff: `f 2 ff`
Fill $02 through $05 with $88: `f 2 88 4`
Fill $02 through $0B with $44: `f 2 44 10`
Fill every other value from $01 through $FF with $6E: `f 1 6e 128 2` (Fill, starting from $01, the value $6e, 128 times, incrementing address by 2 bytes each time)

"f" applies to whichever memory is currently being dumped, meaning it can manipulate VRAM as well as system RAM. It will not write to ROM.

When someone wishes to write a value to multiple locations in VRAM, I believe it's often the case that they want to write that value at some fixed interval. This is typically a power of two, but for troubleshooting/debugging bitmap displays it may be an integer quotient of 640, instead. This was my thinking when adding dec_incr as an optional parameter. That being said, it's entirely optional and may be "a bridge too far" as far as complexity is concerned. Feedback always welcome.

This PR also adjusts the pagedown/pageup behavior of the memory dump to increment/decrement the starting address of the dump by a full page, instead of the previous value which was somewhere around 1/3rd of a page.


indigodarkwolf:
Also, if you're inclined to accept this, feel free to squish this into a single commit, because apparently I can't unless I want to delete my entire fork of the X16 emulator. Apparently, I can't submit the PR in squished form, you'll have to be the one to squish it.

I don't get why Github makes this aspect of collaboration so difficult, or why I can't submit a pull request to the project without my own fork, or why I can't have multiple forks, or why I have to commit a pull request just to get my fork up-to-date with the "main" project.

mist64:
I always squash when merging a pull request, so that's no problem! :)

PR #286, opened 2020-07-11
Fixing VRAM memory dump

indigodarkwolf:
The merge to add zero-page registers to the debugger caused the system RAM dump to always display, regardless of the new memory dump mode. This fixes the display so system RAM and ZP registers are only displayed in the RAM dumpmode, and not in the VERA dumpmode,

### DIFF
--- a/debugger.c
+++ b/debugger.c
@@ -468,10 +468,9 @@ void DEBUGRenderDisplay(int width, int height) {
 
 	DEBUGRenderRegisters();							// Draw register name and values.
 	DEBUGRenderCode(20, currentPC);							// Render 6502 disassembly.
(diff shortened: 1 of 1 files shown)


mist64:
Now we have a conflict. :(

indigodarkwolf:
Whoops, guess I forked this branch from the wrong label and the WAI instruction work was in there, causing a conflict with the added work to put in those other opcodes.

indigodarkwolf:
The conflicts should be resolved now, this should be good to merge.

mist64:
Conflict again!! :( I should have merged the pul requests in a different order. :/

indigodarkwolf:
Well, it should be ready to go again.

Issue #246, opened 2020-03-07
Missing 65c02's WAI instruction


ekmett:
The WAI instruction for the 65c02 can be used for more prompt IRQ response.

It places the CPU in low power mode waiting until an IRQ happens, at which point it immediately wakes up, saving a couple clock cycles. This replaces a the usual spin loop pattern.

More interestingly, if the interrupts are disabled, when the CPU is frozen waiting on a WAI, the IRQ just resumes at the next instruction instead of executing the interrupt handler. This saves all the setup of firing off the IRQ handler, RTI to come back, etc. This can be handy if you wanted to do something tricky and time sensitive with, say, the VERA's IRQ for reaching a given line.

mist64:
Fixed in https://github.com/commanderx16/x16-emulator/pull/287

Issue #177, opened 2019-10-15
BBR & BBS opcodes missing


Oleg-Imanilov:
I can't see opcodes in cpu/65c02.opcodes
Is there any reason why it is not there? More important now $ff - is "dbg" which in conflict with bbs7!

```
; **************************************
;              branch on bit
; **************************************
bbr0 zp 2 $0f
bbr1 zp 2 $1f
bbr2 zp 2 $2f
bbr3 zp 2 $3f
bbr4 zp 2 $4f
bbr5 zp 2 $5f
bbr6 zp 2 $6f
bbr7 zp 2 $7f

bbs0 zp 2 $8f
bbs1 zp 2 $9f
bbs2 zp 2 $af
bbs3 zp 2 $bf
bbs4 zp 2 $cf
bbs5 zp 2 $df
bbs6 zp 2 $ef
bbs7 zp 2 $ff
```

Kobrasadetin:
This needs attention, as some (higher level languge) compilers could output correct 65c02 opcodes that will not work in the emulator.

Pigu-A:
Those instructions (along with RMBx and SMBx) were added later in R65C02 which is ported back later in W65C02 (notice the prefix letters). However, they are not supported in 65C816 since it was extended from the initial 65C02 and should not be emulated since there is a plan to upgrade to that CPU.

mist64:
I wouldn't say there is a *plan* for the 65816, but I think we should stick with the basic set of 65C02 instructions anyway – for now. 

We could implement them in the emulator, but we should put a conditional around the support.

ekmett:
Related: #246

mist64:
Fixed in https://github.com/commanderx16/x16-emulator/pull/287

PR #287, opened 2020-07-13
Adding WAI, BBS, BBR, SMB, and RMB instructions, moving dbg to $DB (issues #246, #177)

indigodarkwolf:
Adding the wait-for-interrupt instruction at opcode $cb, implied addressing only, costing 3 cycles. 

The emulator's CPU will idle while a wait flag is set, updating clock cycle but not doing any other work (no moving the instruction pointer, no executing any instructions, etc). IRQ and NMI clears the wait flag along with their other work. This should emulate the 65c02's wait behavior.


indigodarkwolf:
Added other missing opcodes from the 65C02, which means this will now address issue #246 and issue #177.

mist64:
Excellent, thanks!

PR #290, opened 2020-07-15
Fix VRAM dump in the debugger. Fixes #289

### LINKED ISSUE #289: Register view interferes with VRAM dump in debugger
You can see the "R"s and "x"s in the middle of the VRAM dump view when you use the "v" command.  This may be fine since I'm not sure if it's actually overwriting any VRAM data characters, but it looks kind of awkward.

gaekwad:
Only render zero page registers when dumpmode is DDUMP_RAM.  Also remove duplicate DEBUGRenderData call that was causing it to always be rendered regardless of dumpmode.


gaekwad:
Nevermind this has already been fixed by anopther pull request.

PR #291, opened 2020-07-20
Fix SAVE() to match LOAD()

mjallison42:
Change to SAVE() to match LOAD() only for $A000 - $C000. 

This doesn't match all behavior wrt to other RAM regions, or saving several $A000 banks, like SAVE().


mist64:
e84dc1b changed the hypercall architecture so that the emulator reuses the ROM's logic and only deals with reading/writing bytes. This causes the emulator to behave the same with the host filesystem as it does with the SD card.

PR #293, opened 2020-07-23
Improve directory as SD Card.

dillof:
- Add command line option -sdcard-dir to specify directory to use.
- Convert filenames so lowercase file names show up as unshifted in X16 (as documented in README.md).
- Omit . and .. from directory listing.

### DIFF
--- a/loadsave.c
+++ b/loadsave.c
@@ -17,6 +17,19 @@
 #define MIN(a,b) (((a)<(b))?(a):(b))
 #define MAX(a,b) (((a)>(b))?(a):(b))
 
+char *sdcard_dir;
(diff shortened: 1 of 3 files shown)


ZeroByteOrg:
. and .. are necessary, and the real Kernal routines return these anyway, so the emulator should do likewise.
For instance, Calliope uses the presence/absence of .. as an indication of whether the current directory is the root or a subdirectory.

irmen:
I would say this PR is pretty much made obsolete by #466

dillof:
I agree, that is a more full fledged improvement of hostfs.

PR #296, opened 2020-07-27
Update README to fix debugger op-code change from $FF to $DB

jestin:
This updates the README to reflect a change that was made in [81c38c3](https://github.com/commanderx16/x16-emulator/commit/81c38c3c2ffc70ee02e2b0a4316a6eeee4cb323a) to use a different un-used 6502 op-code as the debug breakpoint op-code.

### DIFF
--- a/README.md
+++ b/README.md
@@ -209,7 +209,7 @@ The debugger keys are similar to the Microsoft Debugger shortcut keys, and work 
 |F12|is used to break back into the debugger. This does not happen if you do not have -debug|
 |TAB|when stopped, or single stepping, hides the debug information when pressed 			|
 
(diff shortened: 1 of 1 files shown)


greg-king5:
Neither op-code is a no-operation **6502** code.  They're **65c02** op-codes.  `$FF` is a branch op-code, and `$DB` is the `STP` code.  `STP` (it means "stop") is the most appropriate code number for triggerring a debugger.

jestin:
Perhaps the README should refer to $DB as "the stop operation"?  This change was really just concerned with making the docs reflect the newly updated op code, but I don't mind updating it for increased accuracy.

mist64:
Thanks!

PR #299, opened 2020-08-02
Adding '-version' shell option

Schol-R-LEA:
Adds a command-line option '-version' which prints the current version information and the current and last prior release notes.


mist64:
I agree the emulator should print its version, but I am not a fan of duplicating the release notes into the binary. With the version number, users can look up the release notes.

Also, you checked in a rom symlink, it seems.

Schol-R-LEA:
That's understandable, I can see why you wouldn't want that in the binary.
As for the symlink, I had thought I had added a .gitignore entry for that, but apparently I hadn't. That's a rather embarrassing oversight, and I can see why that alone would disqualify the pull req as it exists.

Issue #301, opened 2020-08-05
Build breaks with "initialization discards `const` qualifier" error


jestin:
On Ubuntu 18.04, I get the following error when building:

```cc -std=c99 -O3 -Wall -Werror -g -I/usr/include/SDL2 -D_REENTRANT -Iextern/include -Iextern/src -c disasm.c -o disasm.o
disasm.c: In function ‘disasm’:
disasm.c:22:19: error: initialization discards ‘const’ qualifier from pointer target type [-Werror=discarded-qualifiers]
  char *mnemonic = mnemonics[opcode];
                   ^~~~~~~~~
cc1: all warnings being treated as errors
Makefile:57: recipe for target 'disasm.o' failed
make: *** [disasm.o] Error 1
```

This only happens after commit 0604fca.  This seems to be caused by the addition of the `const` keyword to the header defined by `MNEMONICS_DiSASSEM_HEADER` in `cpu/buildtables.py`.  I'm able to fix it either by removing the `const` from the header template or by removing the `-Werror` flag from `CFLAGS` in the Makefile.

SerErris:
confirm the same on Ubuntu 20.04

PR #302, opened 2020-08-05
Add `const` to declaration of `mnemonic` in disasm.c

jestin:
This fixes [issue 301](https://github.com/commanderx16/x16-emulator/issues/301) by making the `mnemonic` pointer a `const` as well.


jestin:
In addition to the highlighting that the Atom package does, I've added a special file type for editing tiles/sprites such that the background syntax highlighting matches the default palette color that corresponds to the hex character:  
https://raw.githubusercontent.com/jestin/vim-xci/master/hex_tile.png

I don't know if I'll get around to allowing the user to select custom palettes, but it's better than nothing, and allows you to do more of the tile design in your editor.

jestin:
This was addressed in commit a02d0b1.

PR #309, opened 2020-08-25
Send only one mouse update per frame

### LINKED ISSUE #308: Laggy mouse movement
On my computer (running Linux Mint), mouse movement is laggy in the emulator, and it takes a short amount of time for the pointer to catch up with its actual location. I believe this is because SDL2 sometimes sends up to 5 or 6 mouse events each frame. If the X16 is not supposed to handle more than one mouse update per frame, the pull request I am going to submit will prevent the emulator from sending more than one update each frame by combining the multiple updates from SDL2, which should fix this issue.

Elektron72:
ps2.c has been modified to not immediately send changes when mouse_move,
mouse_button_down, and mouse_button up are called. Instead, send_mouse_state
is now accessible from outside ps2.c, and it must be called to send the changes.
This allows us to call the other mouse functions for each mouse update received
from SDL2, which can happen multiple times per frame. Once all of the updates
are processed, send_mouse_state is called, which sends the combined updates
to the emulated system. This prevents the system from receiving too many
updates, which can lead to mouse lag, and therefore fixes #308.

### DIFF
--- a/ps2.c
+++ b/ps2.c
@@ -205,14 +205,14 @@ void
 mouse_button_down(int num)
 {
 	buttons |= 1 << num;
-	mouse_send_state();
+	//mouse_send_state();
 }
 
 void
 mouse_button_up(int num)
 {
 	buttons &= (1 << num) ^ 0xff;
-	mouse_send_state();
+	//mouse_send_state();
 }
 
 void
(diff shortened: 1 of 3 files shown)


Elektron72:
Going to create a new pull request after I finish reorganizing my fork of the repository.

PR #317, opened 2021-01-09
Fix LOAD/SAVE hypercall so debugger doesn't break

indigodarkwolf:
Attempting to step-over or step-into a `jsr $ffd5` resulted in the debugger not having the opportunity to pause execution until after whatever instruction followed the `jsr`.  In the case of the step-over, since the debugger looks for a specific address to pause execution and that address was being processed before the debugger could pause execution, execution would simply continue without the debugger pausing appropriately.

By inserting a "continue" at the end of the hypercall's if-block, the debugger gets it chance to pause execution properly.

### DIFF
--- a/main.c
+++ b/main.c
@@ -997,6 +997,7 @@ emulator_loop(void *param)
 			}
 			pc = (RAM[0x100 + sp + 1] | (RAM[0x100 + sp + 2] << 8)) + 1;
 			sp += 2;
+			continue;
 		}
 #endif
 


kktos:
Excellent catch, Stephen !
I must say I was very tempted to do something about that part which seems to me very inelegant.
Something like having all these special processes to be handle on one single "thing", like a Class Bus.... even if we're not using C++...
For instance, I love very much how OneLoneCoder solve this : https://www.youtube.com/watch?v=nViZg02IMQo&list=PLrOv9FMX8xJHqMvSGB_9G9nZZ_4IgteYf
But one step after another ;)
Also, I guess Michael is a busy bee right now with all the hardware changes made on protov2.
btw, @mist64 Michael, if you need some hands to help you on some tasks, feel free to ask.

mist64:
> Also, I guess Michael is a busy bee right now with all the hardware changes made on protov2.
> btw, @mist64 Michael, if you need some hands to help you on some tasks, feel free to ask.

Truth is, while I was waiting for the Proto #2, I decided to work on non-X16 projects. Now I have the working hardware in my hands, and I'm back.

Help? On the emulator side, this is the remaining big task for Proto #2 support:

https://github.com/commanderx16/x16-emulator/issues/333

And here's a cleanup task that would be nice to have before the next release:

https://github.com/commanderx16/x16-emulator/issues/332

PR #322, opened 2021-02-11
Fix VSYNC

Elektron72:
The VSYNC interrupt on the Commander X16 should trigger as soon as the current scanline is offscreen. Currently, the interrupt only triggers once the scan returns to the top of the screen, wasting many lines that could otherwise be spent performing time-sensitive operations. This change to video.c causes the interrupt to trigger at the proper location.

This issue was previously discussed here:
https://www.commanderx16.com/forum/index.php?/topic/348-updating-screen-on-vblank-takes-too-long/

### DIFF
--- a/video.c
+++ b/video.c
@@ -1002,7 +1002,8 @@ video_step(float mhz)
 			render_line(y);
 		}
 		scan_pos_y++;
-		if (scan_pos_y == SCREEN_HEIGHT) {
+		y++;
+		if (y == SCREEN_HEIGHT) {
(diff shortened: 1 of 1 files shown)


video.c
```diff
@@ -1002,25 +1002,25 @@ video_step(float mhz)
 			render_line(y);
 		}
 		scan_pos_y++;
```
greg-king5, line 1004: You should move that line down to where that variable actually is used.

PR #323, opened 2021-02-14
Switch front and back porches

Elektron72:
The back porch should be at the beginning of a frame, and the front porch should be at the end of the frame. Previously, these were reversed; the front porch was being counted at the beginning of the frame rather than at the end.


lmihalkovic:
yeah .. I noticed that one too .. glad someone fixed it :) seems un-intuitive that the back porch comes first, but oh well ..

Issue #332, opened 2021-03-28
Remove NES controller support


mist64:
The X16 hardware has been finalized to have two (four) SNES ports. The emulator still supports NES or SNES controller mappings. The code should be simplified to only support the SNES mapping.

This also means the -joy1/-joy2/-joy3/-joy4 command line arguments should no longer take arguments; their existence means the controller is enabled.

PR #335, opened 2021-03-29
NES gamepads are dead, long live SNES gamepads!

indigodarkwolf:
As requested in #332. Removing code in some places, changing code in others, updating README.md, and the command-line usage text.


indigodarkwolf:
Please consider taking PR #336 instead of this. I'll still leave this open for now, since this was directly on-point with a TODO task item, but it is incompatible with #336 and mooted by it.

mist64:
Yup, happy to take #336.

PR #336, opened 2021-03-29
Cumulative optimizations

indigodarkwolf:
A collection of optimizations ported over from my C++-based fork of the X16 emulator:

* Rewrote joystick implementation (this supersedes the PR at #335):
  * Joysticks can now bind at run-time, because I have a wireless controller and can never, ever remember to turn it on before launching the emulator.
  * Joysticks behavior is now event-driven instead of requiring step function every CPU cycle and extra state polling separate from the SDL event system.
* Rewrote PS2 implementation:
  * Significantly less branching
  * Now only updates immediately before reads and write to the appropriate VIA ports. Big performance win.
* Performance tweaks to VERA implementation:
  * A trivial API change was staring us in the face, allowing us to reduce the calls to video_step() from one per CPU cycle to one per CPU instruction. Also a significant win.
  * Another trivial API change of the same nature for vera_spi_step().

Altogether, warp speed on a Raspberri Pi4 has gone **from 101% to 157%**.
Non-warp speed on a Raspberri Pi4 has gone from **80% to a rock-steady 100%**.

I'm certain my previous experiments with RPi4 pegged performance in the mid-30% range, and I cannot explain the discrepancy in baseline performance between then and now. Go team? Compiler optimizations added at some point? RPi4 update made more resources available? Whatever the case, this still represents a **significant** jump in performance by removing the most significant bottlenecks outside of the hot paths of the VERA and base CPU emulation, and all platforms will benefit from this.


ps2.c
```diff
@@ -200,29 +190,30 @@ mouse_send_state()
 	}
 }
 
-
 void
 mouse_button_down(int num)
 {
 	buttons |= 1 << num;
+	mouse_send_state();
```
greg-king5, line 197: Pull request #321 removed those function calls, for a good reason.  Why did you add them back?

indigodarkwolf: That's essentially a merge error since I completely rewrote ps2.c/ps2.h on my fork. Will fix.

indigodarkwolf: Should be cleaned up now. Sorry about that. And the emulator should be even faster still as a result. 😃

indigodarkwolf:
Changes resolved with latest submits to master.

indigodarkwolf:
With the latest i2c stuff checked in, I'm seeing 76% -> barely eeking out 100% on an RPi4. I'll be looking at the i2c stuff eventually, I'm sure, to see if I can't give it the same treatment as the ps2 emulation. But not for this PR, and probably not for at least a week.

ps2.h
```diff
@@ -2,32 +2,31 @@
...
+void ps2_step(int i, int clocks);
+void ps2_autostep(int i);
 
-// fake mouse
-void mouse_button_down(int num);
-void mouse_button_up(int num);
-void mouse_move(int x, int y);
-void mouse_send_state();
+    // fake mouse
+void    mouse_button_down(int num);
```
mist64, line 26: please no indentation to line up function names

via.c
```diff
@@ -18,12 +19,19 @@
 // VIA#1
 //
 // PA0: PS2KDAT   PS/2 DATA keyboard
+// PB0-2 ROM bank
```
mist64, line 22: where do these outdated lines come from? :)

via.c
```diff
@@ -37,59 +45,55 @@
 
 static uint8_t via1registers[16];
 
-void
-via1_init()
+void via1_init()
```
mist64, line 48: please keep the type in a separate line, like in all other code

video.c
```diff
@@ -990,12 +991,12 @@ render_line(uint16_t y)
...
 
 bool
-video_step(float mhz)
+video_step(float mhz, float steps)
 {
 	uint8_t out_mode = reg_composer[0] & 3;
 
 	bool new_frame = false;
-	float advance = ((out_mode & 2) ? NTSC_PIXEL_FREQ :  VGA_PIXEL_FREQ) / mhz;
+	float advance   = ((out_mode & 2) ? NTSC_PIXEL_FREQ : VGA_PIXEL_FREQ) * steps / mhz;
```
mist64, line 999: no identation of this kind, please. it will keep diffs minimal.

mist64 (changes requested):
looks good, just some nits

mist64:
Never mind, I merged it, and I'll change the nits myself :)

mist64:
amended by changes the commit, new commit: 1c7de3e

indigodarkwolf:
Awesome, thanks!

ps2.h
indigodarkwolf: Will try to keep this in mind in the future. I blame clang-format.

PR #313, opened 2020-12-08
Debugger Enhancements v2

kktos:
### Changes

- change the repo arch to have sub dirs for SRC, OBJ
- add INI file for setup; works in // with CLI parms - "x16emu.ini"
- place the debugger in its own window
- add a true console to debugger
- add some fonts for setup
- add symbols in disasm
- add many bps, up to 10
- add a bunch of commands to the console - shamelessly inspired by AppleWin console

### Commands
- **m [[bank]address]**
display CPU memory [at the given address]. If not specified the current bank will be used.
- **v [address]**
display VIDEO memory [at the given address].
- **e address values....**
set memory at address with values
- **f address value length [increment:1]**
fill CPU/VIDEO memory
- **sh address,length|start:end value0 value1 ...**
search CPU/VIDEO memory in range (addr,len or start:end) for bytes
- **d address**
disasm from address
- **b rom|ram bankNumber**
set RAM or ROM bank
- **r A|X|Y|PC|SP|P|BKA|BKO|VA|VD0|VD1|VCT value** 
set register value
- **sec**
set carry
- **clc**
clear carry
- **sec**
set zero
- **clz**
clear zero
- **sei**
set interrupts disabled
- **cli**
clear interrupts disabled
- **sed**
set Decimal mode
- **cld**
clear Decimal mode
- **seb**
set break
- **clb**
clear break
- **ser**
set reserved/not used
- **clr**
clear reserved/not used
- **sev**
set overflow
- **clv**
clear overflow
- **sen**
set negative
- **cln**
clear negative
- **bpl**
list breakpoints
- **bp address**
add a breakpoint at the specified address
- **bpc bp_number|***
clear a specific breakpoint or all
- **?|help**
display help
- **sym [address|symbol]**
look-up for the address or symbol and display it
- **symclear bank [symbol|*]**
clear the "symbol" or "*" all from the symbol table for specified bank
- **symload [bank] symbolFilename**
load the symbol table [for the specified bank]
- **symdump symbolFilename**
dump in a file the symbol dictionary
- **code**
Toggle the display in full page of code/disasm
- **data**
Toggle the display in full page of data
- **font [Name|ID [Path]]**
Display loaded fonts or Set the debugger font or Load and Set the debugger font
- **romdebug**
toggle ROM debug mode to allow editing
- **info**
display X16 emulator & debugger info

PR #338, opened 2021-03-29
Debugger enhancements V2 - new hardware

kktos:


mist64:
Sorry, I missed this in time, and now we have conflicts again.

I tried to resolve the conflicts myself, but having played with it for a while, I really think this needs to be done in three steps:
1. reorganize the directory structure
2. clean up the source
3. add the new debugger features

Could you do this? :) Ping me at mist64@mac.com once the first patch is ready, so I can apply it immediately, and it does not fall behind any more. The other two will then be less time critical.

mist64:
For reference: documentation is in #313.

mist64:
I moved all sources to src/ and .o files to build/ (to be consistent with the ROM build process) in 61fc63d, so one less step here.

mist64:
So for the remaining changes, I'd like to have them broken up like this:
1. split up debugger.c/.h and move it into a subdirectory, together with disasm.c/.h
2. debugger cleanup & feature work

and independently: cleanup work in the non-debugger code

mist64:
I separated out 1./2. into branch https://github.com/commanderx16/x16-emulator/tree/kktos-debugger-enhancements - maybe it's useful for a start.

PR #345, opened 2021-04-04
Remove optional rom_labels.h from HEADERS. Fixed web assembly compile…

sebastianvog:
Two minor tweaks.
I think rom_labels.h is still optional, right ? So I removed it from the HEADERS as it gets included below in the Makefile if it exists.

The second tweak was for WebAssembly compilation which complained about missing MAX_PATH, I brought back included that used to be in main.c up until recently. Not sure if other targets would need that too, i could make that specific __EMSCRIPTEN__ otherwise.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -2,6 +2,10 @@
 // Copyright (c) 2019 Michael Steil
 // All rights reserved. License: 2-clause BSD
 
+#ifndef __APPLE__
+#define _XOPEN_SOURCE   600
+#define _POSIX_C_SOURCE 1
+#endif
 #include <stdlib.h>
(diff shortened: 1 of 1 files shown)


mist64:
I just fixed the rom_labels.h problem in 61fc63d, but since this commit also moved all source files into a subdirectory, it broke your main.c change. Also, the change may have broken the WebAssembly build in general – I haven't tested it. :/

Can you have a look?

sebastianvog:
Taking a look now

sebastianvog:
I updated the PR, it now only has the include changes. Compiles for WebAssembly target and runs.

PR #355, opened 2021-05-02
Added anti-aliasing to VERA PSG waveforms

TaleTN:
This adds anti-aliasing using polyBLEPs to the pulse/sawtooth waveforms. It also includes anti-aliasing for the triangle waveform, but this is disabled, because it's good enough as it is (IMHO, YMMV).

This will obviously be slower than without anti-aliasing (so maybe it should be optional?), but performance should still be pretty good.

It should probably be tested some more, because although it seems to work fine in isolation, I haven't yet tested it in the actually emulator itself...

BTW, I assume the actual VERA PSG hardware doesn't suffer from aliasing, right?

### DIFF
--- a/src/vera_psg.c
+++ b/src/vera_psg.c
@@ -22,8 +22,10 @@ struct channel {
 	uint8_t  pw;
 	uint8_t  waveform;
 
-	unsigned phase;
 	uint8_t  noiseval;
+	unsigned phase, inv_freq;
+
+	// uint16_t freq_3;
 };
 
 static struct channel channels[16];
@@ -36,6 +38,15 @@ psg_reset(void)
 	memset(channels, 0, sizeof(channels));
 }
 
+static void
+set_freq(struct channel *ch, const unsigned freq)
+{
+	static const unsigned n = 0x7FFFFFFF;
+	ch->freq = freq;
+	ch->inv_freq = freq ? n / freq : n;
+	// ch->freq_3 = (freq << 1) / 3;
+}
+
 void
 psg_writereg(uint8_t reg, uint8_t val)
 {
@@ -45,8 +56,8 @@ psg_writereg(uint8_t reg, uint8_t val)
 	int idx = reg & 3;
 
 	switch (idx) {
-		case 0: channels[ch].freq = (channels[ch].freq & 0xFF00) | val; break;
-		case 1: channels[ch].freq = (channels[ch].freq & 0x00FF) | (val << 8); break;
+		case 0: set_freq(&channels[ch], (channels[ch].freq & 0xFF00) | val); break;
+		case 1: set_freq(&channels[ch], (channels[ch].freq & 0x00FF) | (val << 8)); break;
 		case 2: {
 			channels[ch].right  = (val & 0x80) != 0;
 			channels[ch].left   = (val & 0x40) != 0;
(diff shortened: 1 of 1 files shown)


mist64:
Happy to merge this. What's with the commented-out code though?

Sorry for the late comment.

TaleTN:
> What's with the commented-out code though?

I also implemented anti-aliasing for the triangle waveform. However, aliasing in the triangle is already barely audible, so I think it's not worth the trouble. But if anyone wants to try for themselves, then they would need to uncomment the code.

Anyway, you could leave that code out, or else maybe I could add some macro that enables/disables the code.

mist64:
Can you please format it so that:
* the functions that do the work are actually in the code, not commented out (so the compiler sees them and will spot any build problems in the future)
* the call to the function is guarded by `#if 0 // enable this for blabla`

TaleTN:
Yeah sure, that is: I've moved most of the code to static functions that are always compiled, but I've also added some macros to enable anti-aliasing. The latter is necessary (IMHO), because else the channel struct would include unsed variables, and the set_freq() function would do unnecessary calculations.

TaleTN:
FYI: I've just fixed (amended) the 2nd commit, because I forgot to update the argument order in the poly_blamp() call, oops!

TaleTN:
I've just found another issue with the triangle anti-aliasing, so I will amend the 2nd commit again soon.

akumanatt:
The actual VERA doesn't do any kind of anti-aliasing (see https://github.com/fvdhoef/vera-module/blob/rev4/fpga/source/audio/psg.v#L159) and generates PSG samples at only 48 KHz.

Also this commit breaks building.

TaleTN:
> The actual VERA doesn't do any kind of anti-aliasing (see https://github.com/fvdhoef/vera-module/blob/rev4/fpga/source/audio/psg.v#L159) and generates PSG samples at only 48 KHz.

If that's the case, then I guess it's probably better to not use any anti-aliasing in the emulator as well.

But if the actual VERA is generating digital samples at only 48 kHz, then having no anti-aliasing at all will probably sound pretty bad (i.e. worse than an oldskool PSG), especially at higher frequencies.

mist64:
@fvdhoef?

fvdhoef:
It is as akumanatt says. There is no anti-aliasing. There is some audible aliasing, especially with the triangle waveform. But it doesn't sound too bad in my opinion. There aren't enough resources to implement something more sophisticated.

mist64:
Should we make the feature opt-in then, with a command line argument?

fvdhoef:
Well the emulator would then be better than the real hardware. Not that useful in my opinion.

TaleTN:
> There is some audible aliasing, especially with the triangle waveform. But it doesn't sound too bad in my opinion. There aren't enough resources to implement something more sophisticated.

Do note that, although I haven't heard the actual VERA, I'm not sure if what you are hearing in the triangle actually is aliasing. Because of the limited bit depth the triangle would be rather noisy anyway, which would almost certainly mask any aliasing. Aliasing would likely be most noticable when doing a frequency sweep using a square wave. Also note that the aliasing was bad enough for me to notice it while watching a YouTube video of the emulator, but YMMV.

> Well the emulator would then be better than the real hardware. Not that useful in my opinion.

I would agree.

akumanatt:
I think what is really needed here is a proper resampling from both YM's output buffer and VERA's output buffer to target sample rate. This would reduce additional aliasing.

PR #383, opened 2022-04-01
Make video timings match actual VERA

akumanatt:
According to the VERA source code. It would be weird if 25.0, 25.175 and 25.2 MHz clocks all exist in one.

### DIFF
--- a/src/video.c
+++ b/src/video.c
@@ -35,21 +35,18 @@
 #define NUM_SPRITES 128
 
 // both VGA and NTSC
-#define SCAN_WIDTH 800
 #define SCAN_HEIGHT 525
+#define PIXEL_FREQ 25.0
 
 // VGA
-#define VGA_BACK_PORCH_X 48
-#define VGA_FRONT_PORCH_X 16
-#define	VGA_BACK_PORCH_Y 33
-#define VGA_FRONT_PORCH_Y 10
-#define VGA_PIXEL_FREQ 25.175
+#define VGA_SCAN_WIDTH 800
+#define VGA_Y_OFFSET 0
 
 // NTSC: 262.5 lines per frame, lower field first
-#define NTSC_FRONT_PORCH_X 80
-#define NTSC_BACK_PORCH_Y 23
-#define NTSC_FRONT_PORCH_Y 7
-#define NTSC_PIXEL_FREQ (15.750 * 800 / 1000)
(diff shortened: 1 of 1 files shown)


mist64:
@indigodarkwolf ?

indigodarkwolf:
Looks fairly good to me, and at this point akumanatt is going to be more familiar with the hardware details of the VERA than I am -- as I've been in crunch mode at work during all this time where folks have been picking apart the now-open source VERA. I'm inclined to accept it, especially if it's more accurate to what the VERA hardware is doing.

Before I do, though, question for @akumanatt: Is the VERA actually keeping two counters? I've also pinged Wavicle on the Discord server to ask about this.

Reason I want to know is because if the VERA is using a single counter and then doing something like resetting it on a mode switch, then besides more accurately replicating behavior with a similar approach in the emulator, we'd get some performance wins from only having to test `ntsc_mode` once. The VERA stuff is the most performance-sensitive area of the emulator, after all.

akumanatt:
It actually keeps two separate counters and then select each unit's output to the composer as seen here: https://github.com/fvdhoef/vera-module/blob/rev4/fpga/source/top.v#L1120-L1126. The counters only get reset on reset signal.

indigodarkwolf:
Cool. Thanks for following up on that.

akumanatt:
Also keep in mind that this is still not 100% accurate as I found yet another composite mode intricacies. Such as VBlank signal occurs when the full horizontal counter is 0 but new line signal (which in turn triggers line compare) occurs when it's 270 instead. But this is harder to write and in low priority for me now.

rileymi:
Unsubscribe

On 2 Apr 2022, at 10:02, Stephen Horn ***@***.***> wrote:

﻿

Cool. Thanks for following up on that.

rileymi:
Unsubscribe

On 2 Apr 2022, at 10:02, Stephen Horn ***@***.***> wrote:

﻿

Merged #383<https://github.com/commanderx16/x16-emulator/pull/383> into master.

rileymi:
Unsubscribe

On 2 Apr 2022, at 08:09, Natt Akuma ***@***.***> wrote:

﻿

It actually keeps two separate counters and then select each unit's output to the composer as seen here: https://github.com/fvdhoef/vera-module/blob/rev4/fpga/source/top.v#L1120-L1126. The counters only get reset on reset signal.

PR #394, opened 2022-04-19
Add emulator-specific register for disabling consumption of Ctrl/Cmd keys

mooinglemur:
While writing an application to take full advantage of PS/2 scancodes, I would like to be able to process keystrokes like Ctrl+A and Ctrl+V without the emulator unconditionally intercepting them.

Added the ability to poke a true value into $9FB6 to disable the emulator's behavior of capturing of these keystrokes.

### DIFF
--- a/src/glue.h
+++ b/src/glue.h
@@ -56,6 +56,7 @@ extern bool log_keyboard;
 extern bool log_speed;
 extern echo_mode_t echo_mode;
 extern bool save_on_exit;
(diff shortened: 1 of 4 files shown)


mooinglemur:
I had considered a command line switch for this instead, but then concluded programmatic control was better.  For instance, one may bind Ctrl+R to a useful function, only for the emulator to intercept the keystroke and unceremoniously reset.

PR #395, opened 2022-04-20
Moving register used for disabling emulator ctrl keybinds due to a conflict with the implementation of register 6 in Box16 emulator

mooinglemur:
After discussing with @indigodarkwolf I'm hoping we can move this functionality from $9FB6 to $9FB7.


indigodarkwolf:
Much appreciated. I will be looking to port over the WAV export functionality from Box16.

Issue #402, opened 2022-05-21
-keymap command line argument does not select the correct keymap


megagrump:
`./x16emu -keymap <any keymap>` does not activate the selected keymap. For example: `-keymap it` selects the `de` keymap.

![image](https://user-images.githubusercontent.com/31128870/169640447-e85953bf-b761-4769-919b-eac33a823d27.png)

Tested in r41 (Marrakech) on Linux x86-64.

indigodarkwolf:
Should be fixed with #403, but see new issue #406

PR #403, opened 2022-05-21
synchronize keymaps with KERNAL set

megagrump:
Solves #402. Unsure if `en-us-int` and `en-us-dvo` are correctly named like this, please review.


indigodarkwolf:
Well, it seems like the keyboard mapping names are not strictly following any ISO or IETF standards, but are more "inspired by" standards like ISO 3166-1 and IETF BCP 47, preferring to use an ISO or IETF tag/sub-tag that seems appropriate. Since there isn't such a tag for "international" or "dvorak" (since the standards refer to languages, not keyboard layouts), I'm personally okay with these non-standard labels, especially since trying to invent our own sub-tags might result in us directly conflicting with an ISO or IETF standard.

For the record, it seems like the kernal is identifying keyboard layouts through KLIDs (see also: http://kbdlayout.info/), which are pretty much just abstract hexadecimal numbers and not meaningful names, hence the attempt to map languages to them.

What might be a good idea in the future, though, is to change these labels to exactly match the keyboard type labels from the kernal itself:

- abc/x16
- en-us/int
- en-gb
- sv-se
- de-de
- da-dk
- it-it
- pl-pl
- nb-no
- hu-hu
- es-es
- fi-fi
- pt-br
- cs-cz
- ja-jp
- fr-fr
- de-ch
- en-us/dvo
- et-ee
- fr-be
- en-ca
- is-is
- pt-pt
- hr-hr
- sk-sk
- sl-si
- lv-lv
- lt-lt

indigodarkwolf:
(I wonder if `en-ca`, immediately after `fr-be`, is a typo. It might come from https://github.com/commanderx16/x16-rom/blob/master/keymap/klc/1009%20Canadian%20French.klc)

mist64:
Now that there is the `KEYMAP` basic statement, there is no need any more for this secret API between the emulator and the ROM to set a numeric keymap index. Instead, the emulator should inject the BASIC statement with the string passed from the command line and not know about the set of legal strings.

indigodarkwolf:
Oh hey, that's a good point, too. Let's make an issue to track that piece of work.

PR #405, opened 2022-05-29
New feature: Testbench, running emulator headless for unit testing

stefan-b-jakobsson:
This PR adds a new switch to the emulator ("-testbench").

When enabled, the emulator runs in headless mode. The feature is made for unit testing by a test runner typically made in a high level language such as python.

During testbench mode, the emulator listens on stdin for commands sent by a test runner. Those commands may be used to set the state of the emulator (registers, memory address values) before you run the code that is to be tested.

After the test is completed, the test runner may check the state of the emulator (registers, memory address values). The emulator sends its responses over stdout.

The communication "protocol" between the emulator and the test runner is described at the top of the file testbench.h.

There is also a python example implementing a unit test:
- The file testbench.py defines a class ("X16TestBench") that makes communicating with the emulator a bit easier
- The file test.py is a unit test, checking the result of an addition

This code is quite new, and I'm sure there are still bugs to be found.


src/main.c
```diff
@@ -884,13 +899,13 @@ main(int argc, char **argv)
 	// Available since SDL 2.0.8
 	SDL_SetHint(SDL_HINT_VIDEO_X11_NET_WM_BYPASS_COMPOSITOR, "0");
 #endif
-	SDL_Init(SDL_INIT_VIDEO | SDL_INIT_EVENTS | SDL_INIT_GAMECONTROLLER | SDL_INIT_AUDIO);
+	if (!headless) SDL_Init(SDL_INIT_VIDEO | SDL_INIT_EVENTS | SDL_INIT_GAMECONTROLLER | SDL_INIT_AUDIO);
```
mist64, line 902: would it be possible to move `wav_recorder_set_path()` and `memory_init()` up (or down) and combine these three `if (!headless)` into one?

src/main.c
```diff
@@ -1205,16 +1226,16 @@ emulator_loop(void *param)
 		if (has_serial) {
 			serial_step(clocks);
 		}
-		new_frame |= video_step(MHZ, clocks);
+		if (!headless) new_frame |= video_step(MHZ, clocks);
```
mist64, line 1229: no single-line `if`, please

mist64 (changes requested):
awesome feature; just nits

stefan-b-jakobsson:
@mist64 I think I got all these.

stefan-b-jakobsson:
Thanks, that was quick.

There are a couple of things I need to clean up, especially in the python example.

For instance, even though the emulator is terminated when the python code exits, the emulator subprocess is still active at that time. This generates a warning you don't want to see.

I will probably get back with a fix in a couple of days.

PR #407, opened 2022-05-31
Testbench

stefan-b-jakobsson:
This is a fix to my previous pull request a couple of days ago, that added the testbench function.

There is a change to the stdin/stdout communication between the emulator and the test runner. At present, the emulator doesn't respond in any way when receiving a successful request to set values, for instance the value of a memory address. Only if the request was invalid will it respond with an error code. It would be easier to write robust test runners if the emulator always gave a response. This pull request makes the emulator always responding with "RDY" unless it responds with an actual hex value wanted by the test runner, or "ERR".

Some work has also been done on the python test runner example, and it should now be more robust.

I propose to move this python example code out of the src folder, as it is not really a part of the emulator.

When creating the TestBench object, this PR makes it possible to set emulator options that are applied when the emulator starts. This could be used to load program files from the host file system or to enable warp mode, and so on.

A timeout check is added to the waitReady function. This is so you don't get stuck forever in a test that doesn't end.

A function to load VICE label files has been added. Those files may be created by cl65 with the -Ln option. If you make unit testing part of the build process this is interesting as the test runner may use the labels to find entry points even if they have changed address.

And finally, the test runner now exits cleanly without warnings about open resources.


stefan-b-jakobsson:
@mist64 I think this work is finished now. 

Please let me know if anything needs changing when you have had time to review.

This also fixes issue #412 by removing the superfluous variable "test" in main.c that was left there by mistake.

stefan-b-jakobsson:
Just wanted to add that this is not my original idea. Credits to Yazarchy, DesertFish, and Stephen Horn (hope I didn't miss anyone) who all contributed to this on the Discord asm-dev channel.

mist64:
Sorry this was lying around so long. Happy to merge it now... could you do another coding style pass on it though, please? Thanks!

stefan-b-jakobsson:
I tried to format the testbench.c code as I think you want it. Please let me know if you want more changes.

PR #418, opened 2022-09-06
Move I2C to VIA1 PA0+PA1

stefan-b-jakobsson:
PLEASE NOTE! This pull request should be tested and reviewed before merge.

The purpose of this pull request is to change the emulator to the latest hardware prototypes. 

That hardware uses VIA#1 pins PA0 and PA1 for I2C communication. 

Moreover, the keyboard and mouse is read by the SMC instead of directly by the 65C02 and VIA. The Kernal fetches keyboard and mouse data over I2C.

The changes to the emulator were far from trivial, and there might still be bugs. Before merging this code, it needs to be tested.


ZeroByteOrg:
This is a very important update. Thanks for the hard work, Stefan!

irmen:
I just tested this PR with the kernal that goes with it (https://github.com/commanderx16/x16-rom/pull/338) .
Everything seems to work fine (mouse, keyboard, including the KEYHDL handler routine). 
I haven't tested the mouse very extensively though (only moving it around) as I don't really have programs that otherwise use it.

stefan-b-jakobsson:
Worked fine for me too.

mist64 (approved):
I took the liberty to change spacing a little

PR #422, opened 2022-09-19
Add a `-widescreen` option for stretching the output to 16:9

jestin:
I realized after watching the X16 demo'd at VCF Midwest that a lot of users are going to hook them up to 16:9 monitors rather than 4:3, and not all of these monitors will have the option to render the image at 4:3.  The sad reality is that we should probably expect the average Commander X16 user to view the output in 16:9 rather than the intended 4:3.

With that in mind, I thought that the emulator should have an option for displaying the output in the way that users are most likely going to be viewing it.  This will be useful for application/game developers to view their projects as they will realistically be seen by the users.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -111,6 +111,7 @@ char *gif_path = NULL;
 char *wav_path = NULL;
 uint8_t keymap = 0; // KERNAL's default
 int window_scale = 1;
+double screen_x_scale = 1.0;
 char *scale_quality = "best";
 bool test_init_complete=false;
 bool headless = false;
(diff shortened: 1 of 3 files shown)


jestin:
One caveat I realize is that this change does nothing to capture GIFs in widescreen mode.  Further changes would be required to enable that, and I'm not sure if that'd be the desired result anyways.  Screen scaling seems to have no effect on GIF captures, so I suppose widescreen mode shouldn't either.

mist64:
Great idea!

PR #423, opened 2022-09-20
Handle SDL_GetBasePath() returning null.

xibbert:
SDL_GetBasePath() returns null on systems that don't implement it.

Set to "./" per SDL docs example.

My OS (apparently) doesn't implement SDL_GetBasePath() and the emulator was crashing on startup.

Note: I haven't tested this on Windows, I don't have a way to do so. I don't know whether Windows gets upset about "./".

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -515,10 +515,15 @@ main(int argc, char **argv)
 
 	char *base_path = SDL_GetBasePath();
 
+    if (!base_path) {
+        base_path = SDL_strdup("./");
+    }
+
 	// This causes the emulator to load ROM data from the executable's directory when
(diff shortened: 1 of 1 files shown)


src/main.c
```diff
@@ -515,10 +515,15 @@ main(int argc, char **argv)
 
 	char *base_path = SDL_GetBasePath();
 
+    if (!base_path) {
+        base_path = SDL_strdup("./");
```
mist64, line 519: Are you sure you need to allocate a copy of this string? The following should work:

`base_path = "./"`

This way, you don't have to SDL_free() later.

mist64 (changes requested):

PR #432, opened 2022-10-13
Host FS MACPTR behavior for fast VLOAD

ZeroByteOrg:
Pending Kernal PR#337 modifies the behavior of MACPTR by adding an argument to the API:
C set = use "stream mode" (do not advance the dst pointer during copy from buffer)
C clear = normal mode

The HostFS implementation needs to match this or else programs break, since the Kernal LOAD routine uses this argument.

### DIFF
--- a/src/ieee.c
+++ b/src/ieee.c
@@ -482,7 +482,7 @@ TALK(uint8_t a)
 }
 
 int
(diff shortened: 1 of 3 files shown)


mist64 (approved):

mist64:
I'll merge this at the same time as the ROM change.

PR #435, opened 2022-10-15
Add host support for CMDR-CBMDOS's seekable read/write files

davidgiven:
This adds host filesystem support for M-mode files, which support arbitrary read/writes, and the SD2IEC form of the P command which allows seeking within them.

One thing I'm uncertain about: I had to hack around CR termination of the P command because it takes binary arguments, and terminating the command on a $0d would cause odd results if one of the parameters was $0d.

I feel like the right thing to do here is to change the command parsing stuff to only execute commands on an UNLSN, but I don't know whether that's actually valid. I can't find any formal specification of how DOS commands work. Would this be safe?

### DIFF
--- a/src/ieee.c
+++ b/src/ieee.c
@@ -42,14 +42,14 @@ int dirlist_pos = 0;
 
 typedef struct {
 	char name[80];
+	bool read;
 	bool write;
-	int pos;
-	int size;
 	SDL_RWops *f;
 } channel_t;
 
 channel_t channels[16];
 
+static void cseek(int channel, uint32_t pos);
 
 static int
 create_directory_listing(uint8_t *data)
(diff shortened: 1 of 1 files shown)


mist64:
> I feel like the right thing to do here is to change the command parsing stuff to only execute commands on an UNLSN, but I don't know whether that's actually valid. I can't find any formal specification of how DOS commands work. Would this be safe?

There has never been a formal specification. I tried to compile all sourced into one document here:
  https://www.pagetable.com/?p=1038
but it doesn't answer this question either.

The "correct" behavior here would of course be to match the one in the DOS code in the x16-rom.

davidgiven:
Thanks!

PR #436, opened 2022-10-15
Clear D Flag on IRQ/NMI

### LINKED ISSUE #414: Cpu simulator doesn't clear D flag on irq
It has come to light that the cpu simulation doesn't properly clear the Decimal (bcd) flag on IRQ and possibly other things. This is improved behavior of the 65C02 over the 6502 NMOS version, and the cpu simulator only seems to simulate the nmos variant's behavior here. 

This causes issues:  https://www.commanderx16.com/forum/index.php?/topic/3635-mouse_get-kernal-function-issue/

65c02 decimal flag behavior on irq, nmi, brk, reset: http://www.6502.org/tutorials/decimal_mode.html#5

At least the irq6502 routine should be improved:  https://github.com/commanderx16/x16-emulator/blob/master/src/cpu/fake6502.c#L180  but most likely the nmi6502 as well and maybe others.

LRFLEW:
This is an attempt to fix #414

The code was already clearing the decimal flag on a `BRK` instruction, so this changes it to also clear it on IRQ, NMI, and RESET signals. I also improved the handling of other flags during these signals. According to [section 3.11 of the manual](https://www.westerndesigncenter.com/wdc/documentation/w65c02s.pdf), the B and I flags are set during a reset, so this PR also sets those flags during a reset. With that change to how the B flag is handled, I also updated the NMI to push the status register with the B flag cleared. I had some difficulty determining how the flag is handled on an NMI, with the only resource I could find being [this one for the NES](https://www.nesdev.org/wiki/Status_flags#The_B_flag), so I'm not 100% confident about how it's handled on the 65C02.

### DIFF
--- a/src/cpu/fake6502.c
+++ b/src/cpu/fake6502.c
@@ -171,8 +171,9 @@ static void putvalue(uint16_t saveval) {
 
 void nmi6502() {
     push16(pc);
-    push8(status);
-    status |= FLAG_INTERRUPT;
+    push8(status & ~FLAG_BREAK);
+    setinterrupt();
+    cleardecimal();
     pc = (uint16_t)read6502(0xFFFA) | ((uint16_t)read6502(0xFFFB) << 8);
     waiting = 0;
 }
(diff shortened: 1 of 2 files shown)


LRFLEW:
Rebased on top of #434 to resolve the merge conflict

PR #439, opened 2022-10-23
Debugging improvements

stefan-b-jakobsson:
This PR is to improve debugging in two ways that are somewhat linked to each other.

Currently, all RAM is initialized to zero in the emulator. This is not the case on real hardware, where you must not rely on the values of uninitialized RAM.

From experience, it's easy to write code by mistake that needs some part of RAM being initialized to zero when you use the current emulator.

To make this kind of bug easier to find, I propose the following changes:

- All RAM is initialized to random values on startup, as it would be on hardware
- The option "-wuninit" is added to the emulator; the emulator keeps record of which RAM addresses have been written to, and a warning is printed to stdout if code reads from uninitialized RAM.

### DIFF
--- a/src/memory.c
+++ b/src/memory.c
@@ -23,6 +23,7 @@ uint8_t ROM[ROM_SIZE];
 
 static uint8_t addr_ym = 0;
 
+bool randomizeRAM = false;
 bool reportUninitializedAccess = false;
 bool *RAM_access_flags;
 
@@ -34,18 +35,24 @@ void cpuio_write(uint8_t reg, uint8_t value);
 void
 memory_init()
 {
-	//Initialize RAM array
+	// Initialize RAM array
 	RAM = calloc(RAM_SIZE, sizeof(uint8_t));
 	
-	//Initialize RAM access flag array
-	RAM_access_flags = (bool*) malloc(RAM_SIZE * sizeof(bool));
-	
-	//Randomize all RAM and set all access flags to false
-	time_t t;
-	srand((unsigned)time(&t));
-	for (int i = 0; i < RAM_SIZE; i++){
-		RAM[i] = rand();
-		RAM_access_flags[i]=false;
+	// Randomize all RAM (if option selected)
+	if (randomizeRAM) {
+		time_t t;
+		srand((unsigned)time(&t));
+		for (int i = 0; i < RAM_SIZE; i++) {
+			RAM[i] = rand();
+		}
+	}
+
+	// Initialize RAM access flag array (if option selected)
+	if (reportUninitializedAccess) {
+		RAM_access_flags = (bool*) malloc(RAM_SIZE * sizeof(bool));
+		for (int i = 0; i < RAM_SIZE; i++) {
+			RAM_access_flags[i] = false;
+		}
 	}
 
 	memory_reset();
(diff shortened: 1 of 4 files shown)


src/memory.c
```diff
@@ -30,7 +34,20 @@ void cpuio_write(uint8_t reg, uint8_t value);
 void
 memory_init()
 {
+	//Initialize RAM array
```
mist64, line 37: please add one space after "//"; here and elsewhere

src/memory.c
```diff
@@ -30,7 +34,20 @@ void cpuio_write(uint8_t reg, uint8_t value);
...
 	RAM = calloc(RAM_SIZE, sizeof(uint8_t));
+	
+	//Initialize RAM access flag array
+	RAM_access_flags = (bool*) malloc(RAM_SIZE * sizeof(bool));
+	
+	//Randomize all RAM and set all access flags to false
+	time_t t;
+	srand((unsigned)time(&t));
+	for (int i = 0; i < RAM_SIZE; i++){
```
mist64, line 46: please add one space before "{" to be consistent with existing code

src/memory.c
```diff
@@ -30,7 +34,20 @@ void cpuio_write(uint8_t reg, uint8_t value);
...
+		RAM[i] = rand();
+		RAM_access_flags[i]=false;
```
mist64, line 48: again, for style consistency, surround operators like "=" with spaces

src/memory.c
```diff
@@ -55,6 +76,29 @@ effective_ram_bank()
 
 uint8_t
 read6502(uint16_t address) {
+	//Report access to uninitialized RAM (if option selected)
+	if (reportUninitializedAccess == true) {
```
mist64, line 80: `if (reportUninitializedAccess) {`

src/memory.c
```diff
@@ -95,6 +139,14 @@ real_read6502(uint16_t address, bool debugOn, uint8_t bank)
 void
 write6502(uint16_t address, uint8_t value)
 {
+	//Update RAM access flag
+	if (address < 0xa000) {
```
mist64, line 143: better guard this with `if (reportUninitializedAccess)`; I fear a speed impact otherwise

src/memory.c
```diff
@@ -30,7 +34,20 @@ void cpuio_write(uint8_t reg, uint8_t value);
...
 	RAM = calloc(RAM_SIZE, sizeof(uint8_t));
+	
+	//Initialize RAM access flag array
+	RAM_access_flags = (bool*) malloc(RAM_SIZE * sizeof(bool));
+	
+	//Randomize all RAM and set all access flags to false
+	time_t t;
+	srand((unsigned)time(&t));
+	for (int i = 0; i < RAM_SIZE; i++){
+		RAM[i] = rand();
```
mist64, line 47: I'm not a fan of randomizing RAM. It makes RAM dumps much less readable. Is the uninitialized warning not enough?

If you want to keep it, I'd put it behind a second command line option.

mist64 (changes requested):

mist64:
(Sorry for the coding style nitpicks. I try to keep the code very consistent.)

stefan-b-jakobsson:
I tried to handle your comments in my b6bb6c8.
I think randomized RAM is useful, as it will tell you more realistically if a program will fly on the real thing. But I now put it behind the option "-randram".
Besides code formatting, I made changes so that these debug options do not have a performance impact when not selected.
The commit ae453bb disables the debug message "buffer full...". It was originally disabled in R35, but I managed to have it return in the code you merged during the weekend. It is way too noisy, and should be removed.

stefan-b-jakobsson:
About code formatting. I have no problem with strict formatting rules - I like that code looks consistent. I must improve my skills, but it's hard to learn old dogs sit (or to format code differently than what they are used to) :-)

PR #454, opened 2022-12-01
Update disasm.c

indigodarkwolf:
The WDC datasheet states that a BRK instruction is actually two bytes long, with the CPU skipping the second byte by incrementing the program counter past it. The CPU emulation is already doing this, it was just being disassembled incorrectly by the emulator.


indigodarkwolf:
The failed builds seem to be temporary 404s from the apt-get install process grabbing SDL2 dependencies. I'll try re-running the failed build processes later.

Issue #456, opened 2022-12-08
Middle mouse button not working


ZeroByteOrg:
The mouse driver does not check use the middle mouse button when updating the mouse.

ZeroByteOrg:
Submitting a PR to fix this.

indigodarkwolf:
Excellent. Marking fixed. :3

PR #457, opened 2022-12-08
Fix for issue #456 - middle mouse button

ZeroByteOrg:
Simple fix.
Verified that the middle button state now appears in the first byte of mouse data as it should.


indigodarkwolf:
Thanks for this! :)

PR #460, opened 2023-01-03
add descriptive reason message when dumping system memory file

irmen:
The emulator writes a memory dump file when pressing CTRL/Command + S, and when the program counter reaches $ffff

This PR adds a descriptive message to the console as well so you can tell why exactly it is writing the dump file. This helps diagnosing what's happening, and for me at least, to tell it apart from an unexpected segfault crash/coredump

```
Dumping system memory. Reason: user keyboard request
Dumped system to dump-7.bin.
Dumping system memory. Reason: CPU program counter reached $ffff
Dumped system to dump-8.bin.
```

### DIFF
--- a/src/glue.h
+++ b/src/glue.h
@@ -62,7 +62,7 @@ extern uint8_t keymap;
 extern bool warp_mode;
 extern bool testbench;
 
-extern void machine_dump();
+extern void machine_dump(const char* reason);
 extern void machine_reset();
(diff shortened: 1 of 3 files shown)


indigodarkwolf:
Looks good to me, and seems like a very reasonable inclusion since folks have gotten confused over this detail multiple times.

PR #461, opened 2023-01-07
Make emulator support RAM increments of 8K

JimmyDansbo:
As things were, it was not possible to start the emulator with, for example, 1536KB of memory which would be possible to have in hardware. This change makes it possible to use any multiple of 8K for the amount of RAM

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -548,7 +548,7 @@ main(int argc, char **argv)
 			}
 			int kb = atoi(argv[0]);
 			bool found = false;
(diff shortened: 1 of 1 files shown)


indigodarkwolf:
Simple enough change, but I'm going to suggest that, since it's changing to accept any multiple of 8KB, we don't need the for-loop anymore to search for valid values. Just compare `(kb & 7) == 0`. If true, we have a multiple of 8. If false, we do not.

indigodarkwolf:
Awesome. Works for me.

PR #462, opened 2023-01-14
Set ISR bits upon VSYNC or LINE IRQs irrespective of what is in IEN

mooinglemur:
On real hardware, the ISR bits are set upon their respective events (VSYNC or LINE)  whether or not they are set in IEN.  The emulator had a check for the IEN bit before setting ISR.

The decision whether to assert IRQ does check both IEN and ISR already, so no change was needed there.

Changing the emulator to match real hardware.

Real hardware behavior:
image

Emulator behavior:
image

Behavior of emulator after this change:
image


indigodarkwolf:
Looks good to me. Thanks!

PR #465, opened 2023-01-18
Clear sprite buffer when disabling sprites in VERA

jestin:
I noticed that when a raster line interrupt occurs on a line where a sprite is being drawn, and the the sprites are disable during that interrupt, the sprite continues to draw what was in the buffer for that line.

image

To fix this, we can clear the sprite buffers when sprites are disabled by a vera register write.  This is the same clearing that happens per-line when rendering sprites, except we don't replace the pixels.

Here's what it now looks like with this fix in place, which is how it looks on box16 as well as on real VERA hardware:

image


indigodarkwolf:
Yep, that looks like a reasonable fix. Thanks!

PR #466, opened 2023-01-18
Directory recogition and traversal support for hostfs

mooinglemur:
This PR requires review before accepting.  I have no way to test whether things work as expected in OS X, but I was able to test Windows and Linux.

My rusty C is likely not optimal, and I might be using weird code patterns that have better library functions, but the changes seem to work and I think I was careful enough that leaks do not accumulate.  But there is a bit of new code that touches the filesystem that might have portability concerns.

Changes include:
* fixed MACPTR wrongly slurped in one extra byte at EOF
* added cmdline options -fsroot <dir> and -startin <dir> for "jailing" the hostfs inside of a subdirectory.
* implemented file unlink and rename via `S:FILE` and `R:NEW=OLD`
* implemented chdir/mkdir/rmdir via `CD:...` `MD:...` `RD:...`
* added directory listing-oriented cwd support via `$=C` like in https://github.com/commanderx16/x16-rom/pull/373

### DIFF
--- a/src/ieee.c
+++ b/src/ieee.c
@@ -62,6 +62,7 @@ bool dirlist_eof = true;
 bool dirlist_timestmaps = false;
 DIR *dirlist_dirp;
 char dirlist_wildcard[256];
+char dirlist_type_filter;
 
 const char *blocks_free = "BLOCKS FREE.";
 
@@ -193,6 +194,7 @@ resolve_path(const char *name, bool must_exist)
 	char *ret;
 	DIR *dirp;
 	struct dirent *dp;
+	int i;
 
 	if (tmp == NULL) {
 		set_error(0x70, 0, 0);
@@ -211,7 +213,7 @@ resolve_path(const char *name, bool must_exist)
 		strcpy(tmp+strlen(hostfscwd)+1, name);
 	}
 
-	if (tmp[strlen(tmp)-1] == '*') { // oh goodie, a wildcard
+	if (strchr(tmp,'*') || strchr(tmp,'?')) { // oh goodie, a wildcard
 		// we have to search the directory for the first occurrence
 		// in directory order
 
@@ -235,15 +237,6 @@ resolve_path(const char *name, bool must_exist)
 			c = d+1;
 		}
 
-		d = strchr(c,'*');
-		if (d == NULL) { // This also should never happen
-			free(tmp);
-			set_error(0x62, 0, 0);
-			return NULL;		
-		}
-
-		*d = 0; // chop off at wildcard char
-
 		if (!(dirp = opendir(tmp))) { // Directory couldn't be opened
 			free(tmp);
 			set_error(0x62, 0, 0);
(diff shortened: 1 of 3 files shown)


irmen:
Just piping in to say I've been testing this with several of my programs (image viewer, a game, file based assembler, and the x16 shell)  and many issues they had on host filesystem have been resolved. Most if not all of the functions now work on host fs where they only worked on sdcard image before.

irmen:
Maybe it also should handle the "@:" filename prefix now, i.e. overwrite existing file
Instead it still treats it as regular part of the filename.

mooinglemur:
There is no @: or wildcard handling yet.  I had originally planned that to be in a future PR, but I can go ahead and work on that support.

irmen:
I'm fine either way, this PR is large enough as it is.

mooinglemur:
I feel that the state of this PR is better than what exists in master, and doesn't seem to break anything that already exists, but it does of course need more review and testing to make sure the new features work as expected

indigodarkwolf:
Maybe this speaks to my (potential lack of) attention to style, but stylistically nothing stands out to me as needing to change. 

The use of dynamically allocated strings strikes me, personally, as an unusual thing, but what with the POSIX functions that can't make path length guarantees anyways and have to be called in such a fashion that they return dynamically allocated strings that the caller is responsible for free()'ing, well... I feel all I can do is check that everything appears to be getting cleaned up appropriately, and I didn't spot any memory leaks.

src/ieee.c
```diff
@@ -113,41 +558,180 @@ create_directory_listing(uint8_t *data)
...
+	char *tmp = malloc(strlen(hostfscwd)+1);
+	if (tmp == NULL) {
+		set_error(0x70, 0, 0);
+		return 0;
+	}
+	int i = strlen(hostfscwd);
+	int j = strlen(fsroot_path);
+	strcpy(tmp,hostfscwd);
+
+	for(;;) {
```
indigodarkwolf, line 658: This `for(;;)` "feels" out-of-place to me, mostly because I tend to get bothered by for-loops that are programmed like `while(true)`-loops, and it seems like it would be appropriate to loop on `for(; i >= j-1; --i)`. 

In particular, the fact that this is programmed like an infinite loop leaves me worried about line 673, below, which is `if (!namlen) continue;`, which appears to be an infinite-loop case unless `tmp[i-1]=0;` on line 664 is making a change that would cause us to decrement and continue via the `if` on line 659.

tl;dr: Is line 673 a bug? And if so, would it be more appropriate to make the stopping condition and iteration behavior of the for-loop be more explicit by placing them in the `for()` statement rather than manually decrementing `i` and checking the break condition at the end of the loop?

mooinglemur: Here's the logical flow of that entire for loop with example strings:

`hostfscwd = "//home/whatever"; fsroot_path = "/";`
- i = 15
- j = 1
- first 8 iterations of the for loop, the first if statement matches, `i` gets decremented and the loop restarts
- the 9th iteration, the if statement is false and it continues on
- tmp, which started out as `"//home/whatever"` is now `"//home\x00whatever"`
- i = 7 now, and is not < j
- file_size = 0 ; the file size string output is copied from the regular directory listing code, so we kept this variable even though it's constant.
- namlen is now 8 as tmp+i is "whatever", and is not zero
- we add the path component `"whatever"` to the output of the directory listing and decrement i, continuing the loop
- starting with i=6, we hit true on the first if statement 4 times
- i now = 2, and the if condition is false because tmp[1] is a /
- we make that spot null, so now tmp = `"/\x00home\x00whatever"`
- i is still < j
- namlen is 4
- output `"home"` as path component and decrement i
- starting with i=1 the if is immediately false
- tmp becomes `"\x00\x00home\x00whatever"`
- namlen is now 0, and we repeat the loop without decrement
- this time the first if is true because we clobbered the leading `/`
- i is now 0 and is less than j so tmp becomes `"/"`
- and this is the final path element that gets appended to the data output

tl;dr, namlen is never 0 unless tmp[i-1] is not a path separator character.  And the loop will not get stuck because either the first if will be true, or the second one, if (i < j), is true.

mooinglemur: Come to think of it, all of the `continue`s are either associated with an `i--` or for the `namlen` check, an `i--` would simply save one iteration and not change the important part of the flow.  We can put `i--` in the third arg of `for()`.  As for the check condition, we can also get rid of the `break` with your suggestion.

indigodarkwolf:
Awesome. So many thanks for all the work here!

irmen:
Hurrah !

indigodarkwolf:
There is a coda, which I'll leave here though it should probably be elevated to an issue (maybe a good first-timer issue, even?). To remove some, but not all, of the dynamic allocations of strings, we should consider replacing at least some of the `alloc`s with `alloca`. `alloca` performs a dynamic allocation onto the function stack, meaning you don't need to `free` it later to clean it up. Very useful as long as we're not talking about many kilobytes or later.

irmen:
is alloca still a thing in C++ btw?

indigodarkwolf:
Sure is.

PR #467, opened 2023-01-18
Implement new audio mixing and emulation timing and update VERA audio emulation

akumanatt:
In preparation of a new YM2151 core which lacks resampling, the audio routine is very much rewritten.
- The mixer can mix VERA and YM2151 sources sampled in each of its native sample rate
- The mixer can output in any sample rate, 48828Hz sample rate request is now not enforced to SDL
- Implement a 4-tap filter to resample with minimal aliasing
- Mixing between VERA and YM2151 is changed to reflect Proto3 hardware: VERA is now 6 times louder and YM2151 is 1.5 times louder. A simple limiter is implemented to remove potential clippings from this change.

Timing now uses elapsed CPU clocks as a base instead of elapsed frames, as VERA has different frame rates in each output mode which is not exactly 60Hz and reliable. This reduces (almost) all crackling problems introduced after the video timing rewrite.

VERA PSG and PCM behavior is updated to upstream hardware. Changes including full 24-bit mixing, more PSG resolution (less triangle aliasing and finer volume steps), same LFSR noise algorithm and phase resetting on disabled output. However, this necessitates a removal of anti-aliasing code introduced in #355 as this is not a hardware behavior to begin with.

This still needs lots of testing. I'm still not sure if tiny clicks are completely removed and web builds still work.

### DIFF
--- a/src/audio.c
+++ b/src/audio.c
@@ -3,6 +3,7 @@
 // All rights reserved. License: 2-clause BSD
 
 #include "audio.h"
+#include "glue.h"
 #include "vera_psg.h"
 #include "vera_pcm.h"
 #include "wav_recorder.h"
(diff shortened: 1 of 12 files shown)


mooinglemur:
Very much improved audio quality in my testing on Linux.  The clicking is all but gone. <3

indigodarkwolf:
Well, I've finally found the time to review this. I feel confident that this conforms to all of Michael's programming standards, and that I more-or-less understand everything that went in and why, and I agree with all the changes. Thanks for all this work! I'll give Michael until Monday to veto the merge, and if I don't hear back then I'll go ahead. <3

PR #468, opened 2023-01-19
More Proto4 changes

akumanatt:


indigodarkwolf:
I can't recall where, but I know we had a conversation with folks about whether this was the correct approach to clock stretching, and given how the CPU is _currently_ simulated, this seems correct. I'm not sure if this will be a hazard for @ziplantil and pull request #437, but I'm going to go ahead and accept this PR anyways because it represents known hardware changes. 

@ziplantil, I apologize if it seems like your PR has been left to stew; that was not my intention, and I'm sure it wasn't Michael's intention, either. I am sure I didn't get a notification from Github when it was flipped over to "ready", and by the time I noticed, there was a bunch of other stuff submitted that I knew was being anticipated by folks in the wider community.

PR #469, opened 2023-01-20
CI/CD: Add Win32 build, remove tar step on Linux/Darwin builds

mooinglemur:
There was an unnecessary tar step for Linux and Mac builds in the GitHub workflows. Besides creating what was actually a .tar file while calling it .tar.gz, it was still wrapped in a .zip file by the GitHub CD workflow.

A plain zip file is totally fine even for the Linux and Mac builds.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -37,6 +37,44 @@ jobs:
         with:
           name: X16-Emu Windows 64-bit
           path: emu_binaries/*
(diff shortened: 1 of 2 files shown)


mooinglemur:
added Win32 build as well, figured I might as well wrap it into the same PR

indigodarkwolf:
Nice! Thanks!

PR #471, opened 2023-01-26
Add both of the recent flavors of MacOS to CI/CD

mooinglemur:
Change workflow to build explicit MacOS 11 and 12 builds rather than always using macos-latest

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -102,9 +102,9 @@ jobs:
         with:
           name: X16-Emu Linux 64-bit
           path: emu_binaries/*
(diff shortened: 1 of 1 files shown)


mooinglemur:
Hmm, one of the unrelated workflows failed from what looks like a timeout error.  I think the PR is otherwise fine.  A maintainer of commanderx16/x16-emulator can request that the failed workflow be rerun, but otherwise I think it's fine.

indigodarkwolf:
This is easy to review. Looks fine to me. Michael can always change the name of the output artifact if he cares. :)

PR #473, opened 2023-01-26
SPRCOL bits should be set in ISR irrespective of IEN

mooinglemur:
AFLOW, SPRCOL, LINE, and VSYNC bits in ISR should be set regardless of what happens to be in IEN.  In https://github.com/commanderx16/x16-emulator/pull/462 we fixed this for LINE and VSYNC, but I missed SPRCOL, which is fixed in this PR.

Hardware:
image

Emulator before fix:
image

Emulator after fix:
image

### DIFF
--- a/src/video.c
+++ b/src/video.c
@@ -943,12 +943,10 @@ static void
 update_isr_and_coll(uint16_t y, uint16_t compare)
 {
 	if (y == SCREEN_HEIGHT) {
-		if (ien & 4) {
-			if (sprite_line_collisions != 0) {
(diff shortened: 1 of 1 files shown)


indigodarkwolf:
If only all the PRs were this easy. 😇

indigodarkwolf:
That said, just out of an abundance of caution, I'll be merging the larger audio CL first. Looking at the code, I don't anticipate a conflict, but I've been surprised before. So I look forward to merging this on Monday.

PR #476, opened 2023-02-04
Add WebAssembly build to GitHub Actions (CI/CD)

mooinglemur:
Confirmed that this produces a working build that works in a browser.


indigodarkwolf:
Well, well! I'm surprised the web build still works after this much time without having been maintained. Very cool, and I'm happy to merge that one in right away since I happened to still be awake.

Issue #416, opened 2022-08-19
File operations (OPEN/PRINT#) in current directory fails


SerOlmy16:
This might be the drive emulation in the emulator, or it could be a ROM issue.  Consider the following program:

10 OPEN 1,8,1,"TEST"
20 PRINT#1,"FOO"
30 CLOSE 1

Expected result with no SD card file mounted: The file "TEST" should be created in the current directory, containing the text "FOO".

Actual result: The emulator prints a newline and a question mark to the console, and then locks up. A file "TEST" is indeed created in the current directory, containing only the first character of the string ("F" in this case). The file remains exclusively locked until the emulator is closed.

This is 100% reproducible under 64-bit Windows. I have not tried running the emulator on other platforms.

irmen:
99% likely caused by the current minimalistic host filesystem support in the emulator.

mooinglemur:
I found that this was because emulated hostfs CIOUT did not clear the carry bit in the status flag on return.  Without that, I get `?I ERROR in 20` from BASIC, which is bafflingly confused.  

However the real KERNAL routines always do a `clc` regardless of success, so we need the emulated one to do so unconditonally.

I've added this fix to my hostfs enhancement PR. https://github.com/commanderx16/x16-emulator/pull/466

mooinglemur:
Similar weird behavior also manifested itself on `INPUT#` from hostfs, and the issue was basically the same for that as it was for `PRINT#`, just for reading instead of writing.  Fix in PR#479

PR #479, opened 2023-02-08
HostFS: Fix behavior of realpath() on Windows.  Clear carry flag on return from hostfs ACPTR

mooinglemur:
This PR addresses two distinct problems.

_fullpath() on windows returns a valid path even if the final part does not exist.  This differs from the behavior of realpath(), so I changed the macro into a function call that implements realpath() behavior on Windows.

Emulated ACPTR was leaving the state of the carry flag alone, which confuses BASIC if it's not cleared..  This is the counterpart of the behavior for write that was described in https://github.com/commanderx16/x16-emulator/issues/416 which I had fixed in the main hostfs PR, but for reading instead of writing.  The real ROM code clears carry unconditionally on return from ACPTR, so the hostfs code must as well.

### DIFF
--- a/src/ieee.c
+++ b/src/ieee.c
@@ -28,8 +28,6 @@
 #include "glue.h"
 #ifdef __MINGW32__
 #include <direct.h>
-// realpath doesn't exist on Windows, but _fullpath is similar enough
-#define realpath(N,R) _fullpath((R),(N),PATH_MAX)
 // Windows just has to be different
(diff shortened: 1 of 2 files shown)


mooinglemur:
fleshed out the behavior of Windows realpath() emulation to follow proper behavior if the second arg is not NULL, even though all of the current uses of the function in the project pass NULL in the second arg.

indigodarkwolf:
Looks fine to me. Thanks for the fixes!

